PR #23, opened 2014-11-20
Helper: make maxValuesPerFacet configurable

arollason:
- If any value is sent through in this field it will override the value
  set in the backend
- Setting null or an empty array as the value for this option will cause
  it to return all the filters


vvo:
Hi @arollason, sorry it took us some time to reply to your pull request.

Do you still use this change? Do you still want us to review it?

Thanks.

PR #29, opened 2015-02-03
add an `npm run examples` command

vvo:
it uses a fork of [http-server](https://github.com/nodeapps/http-server) for now, will update when [PR to ecstatic](https://github.com/jesusabdullah/node-ecstatic/pull/127) merged


bobylito:
:+1: :)

PR #31, opened 2015-02-10
Feature/index.search tests

vvo:


test/lib/functionTester.js
```diff
@@ -0,0 +1,35 @@
+module.exports = createTestFunction;
+
+function createTestFunction( shouldBeCalled, times ){
+  if( times > 0 && !shouldBeCalled )
```
vvo: In-place style for https://github.com/algolia/algoliasearch-client-js/blob/master/src/algoliasearch.js does not have spaces after parens. Pure styling issue but still, we must follow in-place style before we agree on how to style.

What do you think?

test/tables/index.search.js
```diff
@@ -1,16 +1,19 @@
-var applicationID = 'index.search.applicationID';
+var functionTester   = require( '../lib/functionTester' );
+
+var applicationID    = 'index.search.applicationID';
 var searchOnlyAPIKey = 'index.search.searchOnlyAPIKey';
-var indexName = 'index.search.searchOnlyAPIKey';
+var indexName        = 'index.search.searchOnlyAPIKey';
```
vvo: same styling issue as raised before: https://github.com/algolia/algoliasearch-client-js/blob/master/src/algoliasearch.js#L37

test/lib/test-xhr-call.js
```diff
@@ -1,58 +1,61 @@
...
+  );
+
+  t.deepEqual(
+    xhr.requestHeaders,
+    opts.call.expectedRequest.headers,
+    'Request headers matches'
   );
 
+
+  opts.call.args.forEach( function doTestCheck( p ){
```
vvo: p stands for param, why not write `param`? We should try to avoid single letter arguments, what do you think?

test/lib/functionTester.js
```diff
@@ -0,0 +1,35 @@
...
+    console.log( "WARNING : Trying to test if it should not be called with times > 0",
+                 ( new Error() ).stack );
+
+  var numberOfCalls = 0;
+
+  var f = function testFunction(){
```
vvo: Is there a reason to store the function in a variable instead of just declaring it?

```js
function testFunction() {

}

testFunction.hasBeenCalled = function() {
}

testFunction.numberOfCalls = function() {
}

testFunction.testFunctionCall = function() {
}

```

test/lib/test-xhr-call.js
bobylito: let's call it arg

test/lib/functionTester.js
bobylito: OK! :) :+1:

vvo:
merged into develop, thanks @bobylito

PR #37, opened 2015-03-04
Request strategy tests + fixes

vvo:
- XHR/XDR uses .timeout. It's always available and sends timeout events
- XDR now reads the status property if any and act accordingly
- try/catch around JSON.parse to suppress uncaught errors
- JSONP fixes:
    - separate onload/onreadystatechange
    - global counter
    - tests on all browsers, OK from IE8 to IE11

Request strategy now well tested.

Maybe some tests missing, but that's enough to be able
to refactor/go V3.


vvo:
@bobylito A lil review? A lot changed but still interesting for you to see.

PR description sums it all.

vvo:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/algolia/algoliasearch-client-js/37)







---

<sup>**[.zuul.yml, line 19 \[r1\]](https://reviewable.io:443/reviews/algolia/algoliasearch-client-js/37#-JjdcA7wWaXhpMt30XMz)** ([raw file](https://github.com/algolia/algoliasearch-client-js/blob/b2df4f5efa5472a610f35f10b0763474ea8c548e/.zuul.yml#L19)):</sup>
This server is started automatically by zuul and we use it for the JSONP tests. Because you cannot intercept/mock JSONP (those are only <scripts>)

---

src/algoliasearch.js
```diff
@@ -120,6 +120,8 @@ var AlgoliaSearch = function(applicationID, apiKey, methodOrOptions, resolveDNS,
     }
 };
 
+AlgoliaSearch.JSONPCounter = 0;
```
bobylito: Could you add a comment to explain this? What about renaming it to JSONP_ID or something that has the semantic of an identifier?

vvo: This will be used as part the JSONP id, so I wonder if this would be better to name it JSONP_ID?

Will add a comment.

vvo:
```js
// This holds the number of JSONP requests done accross clients
// It's used as part of the ?callback=JSONP_$JSONPCounter when we do JSONP requests
```

good?

redox: Go for it!

test/spec/request-strategy/JSONP-creates-script-tags.js
```diff
@@ -0,0 +1,106 @@
...
+  var index = fixture.index;
+
+  fauxJax.install();
+
+  var initialScriptTags = document.getElementsByTagName('script');
+  var initialScriptTagsCount = initialScriptTags.length;
+
+  // check that the current state is clean
+  t.ok(
+    every(initialScriptTags, function noJSONPTag(script) {
```
bobylito: I didn't know we could use lodash functions on HtmlCollection or NodeList, but it seems so and that's pretty good news :)

vvo: Did not know either, I just used it because I thought lodash is smart to understand what is a collection :D

test/spec/request-strategy/JSONP-creates-script-tags.js
```diff
@@ -0,0 +1,106 @@
...
+      if (!script.src ||
+        parse(script.src).pathname !== '/1/indexes/simple-JSONP-response') {
```
bobylito:
```js
return !script.src || parse(script.src).pathname !== '/1/indexes/simple-JSONP-response';
```
It is best to avoid `if(cond) true`

test/spec/request-strategy/JSONP-creates-script-tags.js
```diff
@@ -0,0 +1,106 @@
...
+        postCallbackScriptTags.length,
+        initialScriptTagsCount,
+        'Script tag count back to previous'
+      );
+
+      t.ok(
+        every(postCallbackScriptTags, function noJSONPTag(script) {
+          if (!script.src ||
+            parse(script.src).pathname !== '/1/indexes/simple-JSONP-response') {
+            return true;
```
bobylito: same here

test/spec/request-strategy/JSONP-creates-script-tags.js
vvo: I will make the change but why it is better?

src/algoliasearch.js
bobylito: Good for me too :)

test/spec/request-strategy/JSONP-creates-script-tags.js
vvo: Because: if(true) return true adds redundancy to the code, the return result is already in the condition.

bobylito:
:+1:

vvo:
2015-03-06-101016_896x43_scrot

:shipit:

PR #39, opened 2015-03-06
remove unused callback argument in all make..Request()

vvo:
resulting in fixing IE8 ajax requests

cc @redox @bobylito


vvo:
done in V3

Issue #40, opened 2015-03-06
new AlgoliaSearch(appID, apiKey, opts, ..) => algoliasearch(params)


vvo:
## tl;dr;

```js
var client = algoliasearch(applicationID, apiKey, {
  protocol: 'http', // force using https or http, previously named "method"
  hosts: [], // which hosts to use for all other requests but the first one
  tld: '', // which tld to use when computing default hosts
  timeout: 2000,
  jsonp: null // feature-detect/force/disable JSONP (null/true/false)
});

var helper = algoliasearch.helper(client, indexName, opts);
```

## Boring stuff

The current constructor code has a lot of code going on and can be instanciated with a lot of different options/arity.

Also as it is a constructor, it must be used with `new AlgoliaSearch()`.

I propose to have only one way to instanciate the client, use only named parameters and a factory function instead of a constructor.

This reduces the overhead while using our library.

Current signature is `function(applicationID, apiKey, methodOrOptions, resolveDNS, hosts)` and must be used with new AlgoliaSearch.

New signature will have only named parameters and will be used with `algoliasearch(params)`.

## TODO

- [ ] remove the need of new AlgoliaSearch()
- [ ] remove the need of new AlgoliaSearchHelper()
- [ ] use only named parameters
- [ ] drop `resolveDNS`. It's not used

## References

- named parameters: [hall of api shame: boolean trap](http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html)
- a very good JS api design: [browserify](https://github.com/substack/node-browserify#var-b--browserifyfiles-or-opts) (in case you did not notice, we love browserify)

redox:
I don't think we want a `dsnHost` option, I would prefer a boolean here. If you want to specify a custom hostname you can use the hosts array; isn't it?

vvo:
I agree we can ditch dsnHost and only provide the hosts options (where there's no shuffle, nor tld appending).

But what about the boolean? Do you want a boolean to deactivate DSN when using the default hosts?

redox:
You're right; I think we can just get rid of that boolean; I use it in the Algolia Indices explorer (because I use the JS client to perform indexing operations as well) but that's only because it's prepended even if I specify the hosts.

vvo:
Oh yes, no need for a boolean since as soon as we give hosts: [] custom hosts, we can do any complex case (no dsn, dsn,. ).

allright

Issue #41, opened 2015-03-06
CommonJS/AMD/UMD compatibility


vvo:
## tl;dr;

```js
var algoliasearch = require('algoliasearch');
var client = algoliasearch({
  applicationID: '',
  apiKey: ''
});
```

## Boring stuff

Our current JavaScript client is not commonJS compatible.

It cannot be used with [requireJS](http://requirejs.org/) or [browserify](http://browserify.org/).

The only way to use our module is to copy paste the `<script src=..>` tag in a webpage and use the globally available `AlgoliaSearch`.

This type of JavaScript library usage pattern is old and we must provide modern module loaders compatibility.

## TODO

- [ ] module.exports = algoliasearch; (Why "algoliasearch"? see #40 )
- [ ] algoliasearch.helper = require('./lib/helper');
- [ ] for npm users: no build needed
- [ ] for requirejs/amd/umd/standalone users: build with [browserify standalone](https://github.com/substack/node-browserify#var-b--browserifyfiles-or-opts), `browserify -s algoliasearch index.js`
- [ ] add dist/ to .npmignore
- [ ] stop using grunt for build, browserify handles it all
- [ ] remove browserify-shim from package.json, no more used

PR #48, opened 2015-03-10
new AlgoliaSearch() => algoliasearch()

vvo:
BREAKING CHANGE, changed initialization from `new AlgoliaSearch()` =>
`algoliasearch()` #40
- The only exported global property is now `algoliasearch`
- `opts`:
    - `dsnHost` removed, use `hosts:[dsnHost, other hosts]`
    - `dsn` removed, use `hosts:[non-dsn-host, non-dsn-host]`
    - `requestTimeoutInMs` renamed to `timeout`
    - `method` renamed to `protocol`. Protocol should be specified as
`http:` or `https:`
    - `hosts` behavior changed, when using custom hosts:
      - no shuffling
      - no dsn host is prepended to the list
- removed `AlgoliaExplainResults`, no obvious use ATM
- calling algoliasearch() without an applicationID or apiKey will
throw
- Helper: `new AlgoliaSearchHelper()` => `algoliasearch.helper()`,
same function signature

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,7 @@
+module.exports = algoliasearch;
+
+function algoliasearch(applicationID, apiKey, opts) {
+  var AlgoliaSearch = require('./src/algoliasearch');
+
+  return new AlgoliaSearch(applicationID, apiKey, opts);
+}
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -1,479 +1,483 @@
-(function($) {
-  var extend = function(out) {
-    out = out || {};
-    for (var i = 1; i < arguments.length; i++) {
-      if (!arguments[i]) {
-        continue;
-      }
-      for (var key in arguments[i]) {
-        if (arguments[i].hasOwnProperty(key)) {
-          out[key] = arguments[i][key];
-        }
+module.exports = helper;
+
+/**
+ * Algolia Search Helper providing faceting and disjunctive faceting
+ * @param {AlgoliaSearch} client - An AlgoliaSearch client
+ * @param {string} index - The index name to query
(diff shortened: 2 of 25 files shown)


vvo:
So we are compatible with npm/browserify/commonJS.

Now, onto building with browserify and AMD/global compatibility, see #41

vvo:
cc @bobylito @redox

src/algoliasearch.helper.js
```diff
@@ -1,479 +1,483 @@
...
-    out = out || {};
-    for (var i = 1; i < arguments.length; i++) {
-      if (!arguments[i]) {
-        continue;
-      }
-      for (var key in arguments[i]) {
-        if (arguments[i].hasOwnProperty(key)) {
-          out[key] = arguments[i][key];
-        }
+module.exports = helper;
```
vvo: nothing changed in this helper implementation BUT:
- had to remove the wrapping function, no more needed in commonJS/browserify world

bobylito:
Yeah breaking changes! (/me reads code)

index.js
```diff
@@ -0,0 +1,7 @@
+module.exports = algoliasearch;
+
+function algoliasearch(applicationID, apiKey, opts) {
+  var AlgoliaSearch = require('./src/algoliasearch');
+
+  return new AlgoliaSearch(applicationID, apiKey, opts);
```
bobylito: I usually put the defaults in this kind of function so it separates the code of the constructor from the noise induce by defaults.

vvo: You mean https://github.com/algolia/algoliasearch-client-js/blob/new-usage-40/src/algoliasearch.js#L28-L43?

bobylito: I mean that this code could move to the function that does the new instead of the constructor. This is an idea and not something blocking at all.

bobylito:
LGTM :+1:

PR #49, opened 2015-03-10
commonJS/UMD compatibility

### LINKED ISSUE #41: CommonJS/AMD/UMD compatibility

vvo:
* BREAKING CHANGE, no more window.ALGOLIA_VERSION
      - you can find the version in algoliasearch.version
* UMD compatibiliy #41
    - algoliasearch can now be used with any module loader
    - build is now done with browserify
    - no more grunt
* added simple test on `algoliasearch.helper()`
* removed vendor/
  - vendor/json2.js now as an npm module
  - vendor/jquery.typeahead.js now on cdnjs (examples)

fixes #41


vvo:
cc @redox @bobylito

vvo:
@bobylito with this PR you will be able to `npm install algoaliasearch` then `var algoliasearch = require('algoliasearch')` (only browser for now)

build.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Use this with `npm run build` so that `./node_modules/.bin` are injected into
+# the PATH
+
+license='/*! algoliasearch 2.9.2 | © 2014, 2015 Algolia SAS | github.com/algolia/algoliasearch-client-js */'
```
bobylito: it would a nice to have for contributors to have the version number in properties, env or whatever and only edit one file :)

examples/jquery.html
```diff
@@ -1,10 +1,14 @@
+<!doctype html>
 <html>
   <head>
     <meta content='text/html; charset=utf-8' http-equiv='Content-Type'>
     <meta name="viewport" content="width=device-width,initial-scale=1">
+    <!-- Make example work in IE7 -->
+    <script src="//cdn.jsdelivr.net/console-log/2.2/console.min.js"></script>
```
bobylito: What is the requirement for a console?

build.sh
vvo: Yes I agree, after doing this build script, I thought the same. I hate it.

Maybe I'll go gulp/shelljs or env variable injection, I am sure a module exists, will have a look.

examples/jquery.html
vvo: There's a console.log inside the code, IE7 has no console.

That's why ""

build.sh
bobylito: It doesn't seem there is an easy way to ask npm about the current package. Maybe I'm missing something though :/

examples/jquery.html
bobylito: Yeah but where do we have a console use? (it's not in the changes)

bobylito: OK it's in the fail... Why not then. sorry

bobylito:
LGTM :+1:

redox:
When running `npm run dev`, does it rebuild the `dist` after each file changed? If not, we could use RequireJS to directly load the `src` directory -> it might confuse the users reading the examples however?

vvo:
> When running npm run dev, does it rebuild the dist after each file change?

No more, when using `npm run dev` it justs does a build in memory (zuul) and serve it as the test build. Then when you do a file change, it rebuilds it, in memory.

The dist/ is only updated if you do `npm run build`.

redox:
OK, just wanted to make sure we'll not lose productivity. LGTM :+1:

Issue #45, opened 2015-03-09
Return promises to every call


vvo:
## tl;dr;

```js
index.search('smg')
  .then(function(res) {
    // console.log(res.hits);
  })
  .catch(function(err) {
    console.error(err);
  });
```

## Boring stuff

Current JavaScript API client only sends promises when using Angular.JS or jQuery build.

We must return [promises](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise) to every API command.

We won't use Q or then, we use the native promise API, this is not a taste choice, only more relevant.

## TODO

- [ ] use https://www.npmjs.com/package/promise to return promises to every API command

Issue #46, opened 2015-03-09
Merge Node.js/io.js and browser JavaScript clients


vvo:
## tl;dr;

We currently have two JavaScript-like clients:
- Node.js: [npm/algolia-search](https://www.npmjs.com/package/algolia-search), [algolia/algoliasearch-client-node](https://github.com/algolia/algoliasearch-client-node)
- Browser: [npm/algoliasearch](https://www.npmjs.com/package/algoliasearch), [algolia/algoliasearch-client-js](https://github.com/algolia/algoliasearch-client-js)

Merge them in only one client: `algoliasearch`.

## Boring stuff

We can have [isomorphic](http://isomorphic.net/) JavaScript! Let's use the same JavaScript client on both Node.js and the browser.

As the request strategies are a little bit different between Node.js (keepalive using [node-modules/agentkeepalive](https://github.com/node-modules/agentkeepalive)) and the browser (fallback to JSONP, non relevant for Node.js), there will be some code differences.

Luckily, using browserify we are able to swap out requires when in the browser.

### First steps
- [x] Create V~~2~~1 branch in the Node.JS client, patches will be done in this branch
- [x] first prototype exposing AlgoliaSearch JavaScript client in Node.JS
- [x] make testing stack handles the Node.JS client

### Behavior
- [x] ~~All indexing requests must use a timeout of 15 times the requestTimeout given (so 2s search timeout, 30s indexing timeout)~~
- [x] Set default timeout to 3s. Timeout is an inactivity timeout as per [Node.JS request.setTimeout](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback)
- [x] All indexing requests must not use the dsnHost in both browser and node
- [x] Ensure headers must not be mutable between two requests calls (add a test)
- [x] Keepalive should be activated by default, it will not be possible to give another agent for now. It's not a requested feature, it should just works™
- [x] send x-user-agent as part of the url containing "Algolia for Node.JS/io.jsNode.js $NODEJSVERSION $MODULEVERSION" in both browser and node
- [x] let the user use only http if he wants to
- [x] remove any reference to document/window in AlgoliaSearch.js
- [x] provide a .destroy method in nodejs that will destroy the http Agent (and exiting the process)
- [x] implement custom headers: _addHeadersRateLimit, _addHeadersSecuredAPIKey

### Methods
- [x] **Implement** some Node.JS client methods not implemented in the JavaScript client
 - [x] index.getObjects
[...]

PR #52, opened 2015-03-18
Externalize plugins

### LINKED ISSUE #44: Plugins architecture
## tl;dr;

There's no plugin infrastructure in the code, resulting in `if (angular)` `if (jquery)`.

Create one that allows us to remove the specific builds code from the main `algoliasearch.js`.

## TODO

- [ ] create a plugin way to create specific jQuery and Angular builds rather than storing the custom code in the main `algoliasearch.js` file.
- [ ] Remove specific builds code from main algoliasearch.js

## Boring stuff

We have [jQuery and Angular.js builds](https://github.com/algolia/algoliasearch-client-js/blob/4d28d22e7726b537c160190f337e2cafd68b0d63/README.md#jquery-angularjs) but the code needed by theses plugins is stored [inside the main algoliasearch.js](https://github.com/algolia/algoliasearch-client-js/blob/4d28d22e7726b537c160190f337e2cafd68b0d63/src/algoliasearch.js#L673) file.

This is not good because once we have 10 plugins we will be screwed and it complexifies the code a lot.

We must find a way to plug theses builds

We must have one entrypoint per plugin like today: [src/algoliasearch.angular.js](https://github.com/algolia/algoliasearch-client-js/blob/4d28d22e7726b537c160190f337e2cafd68b0d63/src/algoliasearch.angular.js) but it should store his own custom code.

vvo:
* Externalize plugins and request implementations
* NEW FEATURE, All calls are now returning promises.
 - If there's a callback given to an API call, you won't get a promise.
 - Here are the different promises implementations we use:
     - Native promises by default (https://github.com/jakearchibald/es6-promise/)
     - AngularJS promises for AngularJS plugin
     - jQuery promises for jQuery plugin

fixes #44 #45

### DIFF
--- a/.zuul.yml
+++ b/.zuul.yml
@@ -19,6 +19,7 @@ browsers:
 server: ./test/support-server/index.js
 scripts:
   - /node_modules/jquery/dist/jquery.min.js
+  - /node_modules/jQuery-ajaxTransport-XDomainRequest/jquery.xdomainrequest.min.js
   - /node_modules/angular/angular.min.js
 html: ./test/template.html
 browserify:
--- a/src/create-algoliasearch.js
+++ b/src/create-algoliasearch.js
@@ -1,0 +1,17 @@
+// this file is a `factory of algoliasearch()`
+// Given a `request` param, it will provide you an AlgoliaSearch client
+// using this particular request
+module.exports = createAlgoliasearch;
+
+function createAlgoliasearch(request) {
+  function algoliasearch(applicationID, apiKey, opts) {
+    var AlgoliaSearch = require('./algoliasearch');
+
+    return new AlgoliaSearch(applicationID, apiKey, opts, request);
+  }
+
+  algoliasearch.version = require('../package.json').version;
+  algoliasearch.helper = require('./algoliasearch.helper');
+
+  return algoliasearch;
+}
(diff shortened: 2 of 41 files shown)


.zuul.yml
```diff
@@ -19,6 +19,7 @@ browsers:
 server: ./test/support-server/index.js
 scripts:
   - /node_modules/jquery/dist/jquery.min.js
+  - /node_modules/jQuery-ajaxTransport-XDomainRequest/jquery.xdomainrequest.min.js
```
vvo: this allows testing the jQuery plugin in IE9, this will be added to the changelog and migration guide

vvo:
The idea behind this PR is to externalize every plugin implementation, thus preparing the work towards #46: merging JavaScript clients.

It also remove complexity and code paths from the main file.

Finally, it provides a double callback() and promises library not just for jQuery and AngularJS plugins but also for regular users, using native promises.

vvo:
@redox @bobylito review!?

src/create-algoliasearch.js
```diff
@@ -0,0 +1,17 @@
+// this file is a `factory of algoliasearch()`
+// Given a `request` param, it will provide you an AlgoliaSearch client
+// using this particular request
+module.exports = createAlgoliasearch;
+
+function createAlgoliasearch(request) {
+  function algoliasearch(applicationID, apiKey, opts) {
+    var AlgoliaSearch = require('./algoliasearch');
+
+    return new AlgoliaSearch(applicationID, apiKey, opts, request);
```
bobylito: style : prefer parameters for partial application in the first place

PR #62, opened 2015-03-20
Callback always last

### LINKED ISSUE #43: ensure all methods accept a callback as last argument and only in last
## tl;dr;

Example:
`index.search` signature is cryptic and embeds strange parameters like delay. It does not makes it clear that we must have a callback in last parameter.

Other signatures does not always have callback as last parameter.

## TODO

- [x] remove delay parameter in .search(), should be implemented by the user if needed
    easily implemented with [lodash.debounce](https://lodash.com/docs#debounce) for example
- [x] search can be called with (query, params) or (params)
- [x] check all method signatures are using a callback as last parameter
- [x] rework addUserKeyWithValidity signature

vvo:
This PR changes all function signatures that were not following the "callback last parameter" defacto convention.

fixes #43


redox:
Nice one. :+1:

PR #64, opened 2015-03-23
load algoliasearch V2 when jsDelivr `latest` detected

### LINKED ISSUE #63: Do not break any client latest/algoliasearch.min.js usage
## tl;dr;

We need to make the V3 release as smooth as possible in all cases. Even when the user includes the auto-updated "latest" algoliasearch.js

## Boring stuff

We advocated for some time to use http://cdn.jsdelivr.net/algoliasearch/latest/algoliasearch.js which means "Use the latest alogliasearch library which udpates automatically, YOLO".

We want our users to be up-to-date but using versions strings like V2.x V2.9x V3.x etc.. So the user is able to get patch updates and new features updates while being sure his website will always work.

For now we must make the V3 release backward compatible in some way.

## TODO

- [ ] [detect](https://github.com/jehoshua02/getCurrentScript/blob/a391d82375b1adc4338f1ea22cb25a3bd60eb2cd/src/index.js) the algolia script was included using the "latest" version string using jsdelivr (cdnjs does not have such feature)
- [ ] document.write() v2. script (drawback: not very efficient, but temporary)
- [ ] ~~or make a compatibility layer that use the new client from the previous constructor~~ (too complex)
- [ ] in all cases console && console.warn && console.warn('Warning, you are using the latest/ version which is not recommended. See our upgrade guide..') 
- [ ] find all clients using latest/ version and advise them to use jsdelivr V3 and upgrade guide

vvo:
This will make the V3 update handle the `latest/algoliasearch.min.js`
usage.

fixes #63 

@redox @bobylito 

When manually upgrading to V3 without modifying usage:
2015-03-23-151224_1056x46_scrot

When using `latest` version tag from jsdelivr:
2015-03-23-151209_1091x188_scrot

### DIFF
--- a/src/algoliasearch.angular.js
+++ b/src/algoliasearch.angular.js
@@ -86,3 +86,5 @@ global.angular.module('algoliasearch', [])
       }
     };
   }]);
+
+require('./migration-layer')('algoliasearch.angular');
--- a/src/algoliasearch.jquery.js
+++ b/src/algoliasearch.jquery.js
@@ -71,3 +71,5 @@ request.delay = function(ms) {
     }, ms);
   }).promise();
 };
+
+require('./migration-layer')('algoliasearch.jquery');
--- a/src/browser.js
+++ b/src/browser.js
@@ -138,20 +138,4 @@ request.delay = function(ms) {
   });
 };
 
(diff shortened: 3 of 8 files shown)


redox:
Super nice. :+1:

speedblue:
Love it ! :)

vvo:
credits to @redox who thought about the doc.write idea

PR #74, opened 2015-04-01
Fix #73 : if protocol is not http or https, default to http

### LINKED ISSUE #73: In local env, the client uses file protocol
And obviously it is not the expected behaviour, so when the lib tries to uses a protocol different from http or https, we should default to http.

bobylito:


bobylito:
Argh I pointed that to master. Sry.

bobylito:
Creating a new PR.

PR #78, opened 2015-04-02
basic Node.JS client

vvo:
basic node.js client working, still work to do see #46 

Merging this since it will not break anything and a lot of files were moved around so
better not wait too much before merging it.


redox:
:+1:

PR #81, opened 2015-04-18
Working Parse module

vvo:
We now have a working Parse module with the new API

### DIFF
--- a/scripts/parse-build.webpack.config.js
+++ b/scripts/parse-build.webpack.config.js
@@ -15,9 +15,6 @@ module.exports = {
       'util-deprecate': 'util-deprecate/browser.js'
     }
   },
-  node: {
-    process: 'mock'
-  },
   module: {
     loaders: [{
       test: /\.json$/,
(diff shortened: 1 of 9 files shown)


vvo:
Parse cloud locked themselves in a world where you can do JavaScript on the server but it's not Node.js, not iojs, there's no npm compatibility. NOT GOOD.

But it works

vvo:
cc @redox woohoo :)

scripts/parse-build.webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+  entry: path.join(__dirname, '..', 'src/server/builds/parse.js'),
+  output: {
+      path: path.join(__dirname, '..', 'dist'),
+      filename: 'algoliasearch.parse.js',
+      libraryTarget: 'commonjs2'
+  },
+  target: 'node',
+  resolve: {
+    alias: {
+      debug: 'debug/browser.js',
```
vvo: this is where we fallback to browser implems for some requires, so that we get the simple code instead of full nodejs with core requires that will fail in parse.

Parse is like the IE6 of server side JavaScript, easier to get things done this way

redox:
Ahah, well done @vvo !

PR #83, opened 2015-05-01
Syntax errors on examples in README.md

mekza:
According to https://www.algolia.com/doc/javascript#MultiQuery

### DIFF
--- a/README.md
+++ b/README.md
@@ -753,18 +753,21 @@ var client = algoliasearch('ApplicationID', 'apiKey');
 
 var queries = [{
   indexName: 'categories',
-  query: 'search in categories index', {
+  query: 'search in categories index', 
+  params: {
     hitsPerPage: 3
   }
 }, {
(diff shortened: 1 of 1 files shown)


vvo:
Hi thanks for this. Both syntaxes should work, I assume it did not work for you?

mekza:
It didn't work for me, I get  `Uncaught SyntaxError: Unexpected token {`

vvo:
oh yeah sorry it was a syntax error. I first thought you proposed to use {params: query}, it would also have worked.

We have a dedicated project for all api docs: https://github.com/algolia/algoliasearch-client-readme-generator/blob/master/include/JavaScript/multiple_queries.snippet

either we will do the change here or if you are up to, please do it ;)

thanks a lot! Hope you like our new JavaScript client.

mekza:
I didn't see this dedicated repository, I'm doing changes in this one.

Issue #89, opened 2015-05-15
server/builds/node.js: malformed end response crashes app


odeke-em:
Thank you for this cool search service/client!
 
My app's integration tests have been crashing quite a bit whenever deletion of a non-existant index occured. The crash happened because src/server/builds/node.js expects data as parseable JSON
which you can see from:
```javascript 
        // algoliasearch-client-js/src/server/builds/node.js#L135-140
        resolve({
          statusCode: res.statusCode,
          body: JSON.parse(Buffer.concat(chunks))
        });
```

but it actually received back response
```html
<html>
<head><title>400 Bad Request</title></head>
<body bgcolor="white">
<center><h1>400 Bad Request</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

odeke-em:
If you are interested in the client side fix, I have submitted a PR https://github.com/algolia/algoliasearch-client-js/pull/90 to address this.

vvo:
@odeke-em I do agree we should fail-safe bad JSON responses from our own servers and I will fix this behavior.

But I could not reproduce the 400 bad request issue, here's what I tried:

```js
var client = algoliasearch('Good APPID', 'Good Admin Key');
client.deleteIndex('dsadsadkjasfdsfdsfsdfdsfsdfdsfsdfdkdjsa', function(err) {
  console.log(err);
});
```

And I do not receive any 400 bad request, could you double check?

odeke-em:
Hello @vvo, I see that can't reproduce it. Probably it is a mix of issues + maybe rate exhaustion but I certainly caught the errored response and found it was the one tripping things out.

odeke-em:
I can reproduce this using my app btw, just did right now. Sent you some screenshots via email.

vvo:
Great, as a quick fix, can you `npm install algolia/algoliasearch-client-js#98c71478f872a0088fc3b5d02c87ed2876d51d6a` and let me know what happens, you should get an error about JSON parsing.

We are on the problem, cc @Nagriar

odeke-em:
Oh yes, thanks it works great now, here is the output
```
{ 
        [AlgoliaSearchUnparsableJSONError: Could not parse the incoming response as JSON, see err.more for details]
         name: 'AlgoliaSearchUnparsableJSONError',
         message: 'Could not parse the incoming response as JSON, see err.more for details',
[...]

PR #90, opened 2015-05-15
server/builds/nodejs: try-catch end JSON response

odeke-em:
This PR addresses issue #89 on receiving a malformed response that is un-JSON parseable.


odeke-em:
Thank you @vvo for merging this in and for fixing the issue!

vvo:
This is now released, you can `npm install algoliasearch@3.4.0` --save to test it

Issue #86, opened 2015-05-13
Remove console.log calls


renancouto:
Hi, how can I remove all console.log calls on the browser version (I'm using v3)? I couldn't find anything about this on the docs.
Thanks!

vvo:
Hi @renancouto can you show me some of the console log calls you are seeing? Thanks

renancouto:
Hello @vvo! Yes, the logs looks like these:

On init:
```
algoliasearch init done { settings... }
```

On searches:
```
algoliasearch:/1/indexes/*/queries request start +1m
algoliasearch:/1/indexes/*/queries received response: { body... }
```

Don't get me wrong, these logs are nice, but I'd like to removed them in production. Is that possible?

PR #91, opened 2015-05-17
Remove console.log calls

jbpionnier:
disable debug #86


vvo:
thx, I will do a publish tomorrow morning that will deactivate this debug in production. sorry abt that

PR #107, opened 2015-06-01
feat: new browse()/browseFrom()/browseAll()

### LINKED ISSUE #101: Implement new browse

vvo:
- `browse(query, queryParameters)` now has the same signature than
search(). You can use any `query` and `queryParameters`.
- `browseFrom(cursor)` can be used as an efficient way to
continue (next page) a previous `browse()` call. All browse responses now have a `cursor` property.
- `browseAll(query, queryParameters)` can be used to get all
the content of your index
  It returns an [EventEmitter](https://nodejs.org/api/events.html).

  Available events:
    - `result`
    - `end`
    - `error` (you should listen to it or it will throw)

  There's also `stop()` method on the event emitter so that you can
  stop browsing at any point.

fixes #101

### DIFF
--- a/src/AlgoliaSearch.js
+++ b/src/AlgoliaSearch.js
@@ -716,7 +716,7 @@ AlgoliaSearch.prototype = {
           // If there was a JSON.parse() error then body is null and it fails
           httpResponse && httpResponse.body && 200;
 
-        requestDebug('received response: response statusCode: %s, computed status: %d, headers: %j',
+        requestDebug('received response: statusCode: %s, computed statusCode: %d, headers: %j',
           httpResponse.statusCode, status, httpResponse.headers);
 
         if (process.env.DEBUG && process.env.DEBUG.indexOf('debugBody') !== -1) {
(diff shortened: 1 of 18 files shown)


vvo:
review @kokliKo @pixelastic (random!)

src/AlgoliaSearch.js
```diff
@@ -1334,34 +1334,204 @@ AlgoliaSearch.prototype.Index.prototype = {
...
-      params += '&hitsPerPage=' + hitsPerPage;
+    // we check variadic calls that are not the one defined
+    // .browse()/.browse(fn)
+    // => page = 0
+    if (arguments.length === 0 || arguments.length === 1 && typeof arguments[0] === 'function') {
+      page = 0;
+      callback = arguments[0];
+      query = undefined;
+    } else if (typeof arguments[0] === 'number') {
+      // .browse(2)/.browse(2, 10)/.browse(2, fn)/.browse(2, 10, fn)
```
pixelastic: Do you really handle the `.browser(2, 10, fn)` case ? I don't see it in the code path.

src/AlgoliaSearch.js
```diff
@@ -1334,34 +1334,204 @@ AlgoliaSearch.prototype.Index.prototype = {
...
+      queryParameters = undefined;
+    } else if (typeof arguments[0] === 'object') {
+      // .browse(queryParameters)/.browse(queryParameters, cb)
+      if (typeof arguments[1] === 'function') {
+        callback = arguments[1];
+      }
+      queryParameters = arguments[0];
+      query = undefined;
+    } else if (typeof arguments[0] === 'string' && typeof arguments[1] === 'function') {
+      // .browse(query, cb)
```
pixelastic: Shouldn't we also have a `query = arguments[0]` here ?

src/AlgoliaSearch.js
vvo: Yes just below this line there's the case for it and also a test-case that passed: https://github.com/algolia/algoliasearch-client-js/blob/e71939759475bb3c5e0689c821cc24a12c8869f2/test/spec/common/index/test-cases/browse.js#L17-L31

src/AlgoliaSearch.js
vvo: It's already the case since the signature of the function is `function(query, queryParameters)`

I should also add that the order you do:

```js
      callback = arguments[0];
      query = undefined;
```

Is important since callback and query are references to arguments.

So if you do:

```js
      query = undefined;
      callback = arguments[0];
```

In every JS code this would work but here arguments[0] will be undefined as soon as query = undefined is executed.

DOH!JS!

src/IndexBrowser.js
```diff
@@ -0,0 +1,37 @@
...
+IndexBrowser.prototype._result = function(content) {
+  this.emit('result', content);
+};
+
+IndexBrowser.prototype._clean = function() {
+  this.removeAllListeners('stop');
+  this.removeAllListeners('end');
+  this.removeAllListeners('error');
+  this.removeAllListeners('result');
+};
```
pixelastic: Love it. It's clean and understandable quickly.

vvo: Cooool thanks. I was unsure I needed a separate small module to handle this but in the end it's worth it.

src/AlgoliaSearch.js
pixelastic: Ok, didn't clearly see the signature. Makes sense.

src/AlgoliaSearch.js
pixelastic: Oh, yes, ok, I just saw that it's the method signature that defines the `callback`.

PR #110, opened 2015-06-04
switch to lodash-compat and uglifyjs

vvo:
# lodash-compat?
So that we will share more code with the helper and the upcoming
places client.

The build size goes up from 14kb to 17kb.

# uglifyjs

Because, JAVA.


pixelastic:
Agreed. :+1:

redox:
:+1:

PR #115, opened 2015-06-17
fix: parse test must call the api multiple times to avoid cache

vvo:


test/parse/index.js
```diff
@@ -7,33 +7,41 @@ var test = require('tape');
...
-  function done(err, res) {
-    t.error(err, 'no error while contacting our newly deployed parse application');
-    t.equal(res.status, 200, 'status code is 200');
-    t.deepEqual(
-      res.body, {
-        result: expectedResponse
-      },
-      'response was ' + expectedResponse
-    );
+  // sometimes parse cloud sends cached responses, we can avoid
```
pixelastic: Do you feel confident in this fix? Trying 5 times until it works seems pretty weak :)

Is there no other way to bypass the cache?

vvo: > Do you feel confident in this fix? Trying 5 times until it works seems pretty weak :)
> 
> Is there no other way to bypass the cache?

I really dunno, we ask for parse cloud to "deploy" and then it's really deployed "at some point", even if it says that it was deployed, it is not instant.

That's why I added this little retry strategy, not very clean but still working.

Maybe there's a better way, I am not sure it's worth investigating.

PR #121, opened 2015-07-08
chore: pass eslint and jsfmt on whole code

vvo:


bobylito:
Let me have a look :)

.eslintrc
```diff
@@ -1,90 +1,9 @@
...
-  "space-in-parens": [
-    2,
-    "never"
+  "extends": [
+    "eslint-config-airbnb",
+    "eslint-config-algolia"
   ],
-  "globals": {}
+  "rules": {
+    "strict": 0
```
bobylito: Why not use strict?

vvo: I am not sure it brings me anything since eslint already detect unused variables and undeclared global variables.

So since it does not brings me anything but one more line per file, I prefer not to use it. Also 'use strict'; will at some point turn into 'use really-strict' etc..

I think we already had the discussion at some point early on.

src/AlgoliaSearch.js
```diff
@@ -118,54 +119,56 @@ AlgoliaSearch.prototype = {
    *  error: null or Error('message')
    *  content: the server answer that contains the task ID
    */
-  deleteIndex: function(indexName, callback) {
-    return this._jsonRequest({ method: 'DELETE',
-              url: '/1/indexes/' + encodeURIComponent(indexName),
-              hostType: 'write',
-              callback: callback });
+  deleteIndex: function deleteIndex(indexName, callback) {
```
bobylito: This could be a rule to enforce (force naming of the methods) http://eslint.org/docs/rules/func-names ?

src/AlgoliaSearch.js
```diff
@@ -118,54 +119,56 @@ AlgoliaSearch.prototype = {
...
-    return this._jsonRequest({ method: 'POST',
-              url: '/1/indexes/' + encodeURIComponent(srcIndexName) + '/operation',
-              body: postObj,
-              hostType: 'write',
-              callback: callback });
+    return this._jsonRequest({method: 'POST',
+      url: '/1/indexes/' + encodeURIComponent(srcIndexName) + '/operation',
+      body: postObj,
+      hostType: 'write',
+    callback: callback});
```
bobylito: bad indentation

.eslintrc
vvo: I read http://www.nczonline.net/blog/2012/03/13/its-time-to-start-using-javascript-strict-mode/ and it may be interesting for the `this` coercion thing but again not sure adding 'use strict' in every file is worth it given the number of times this happened to me

src/AlgoliaSearch.js
```diff
@@ -177,10 +180,10 @@ AlgoliaSearch.prototype = {
       length = 10;
     }
 
-    return this._jsonRequest({ method: 'GET',
-              url: '/1/logs?offset=' + offset + '&length=' + length,
-              hostType: 'read',
-              callback: callback });
+    return this._jsonRequest({method: 'GET',
```
bobylito: why don't we start the first property after a first carriage return.

src/AlgoliaSearch.js
vvo: why bad? can you rewrite it?

On this subject, there's no ESLint rule I found that solved this or enforced this, so I have left it to every developer preference.

It's being rewritten here by jsfmt who does only indents of two spaces, without aligning (fine with me)

src/AlgoliaSearch.js
vvo: Yes in our eslint-config I have set it as a warning only since it requires some good rewrite and again not everyone likes to do this.

We can set this as an error if we want, still not convinced even if I did it, mostly because it was a warning.

There's no clear benefit of doing it, most recent browsers will show the property name as the function name, so we might even disable it.

what do you think?

src/AlgoliaSearch.js
vvo: we can, I can add the rule for this in ESLint and esformatter, will do

src/AlgoliaSearch.js
bobylito:
```javascript
return this._jsonRequest({method: 'POST',
    url: '/1/indexes/' + encodeURIComponent(srcIndexName) + '/operation',
    body: postObj,
    hostType: 'write',
    callback: callback});
```
the last line doesn't have the same indentation as the other properties. I use indentation to see all the properties easily in objects so I thought that this was a mistake.

vvo: Oh yes you are right this is a mistake, strange it was rewritten this way, will see what I can do

src/AlgoliaSearch.js
```diff
@@ -585,18 +588,29 @@ AlgoliaSearch.prototype = {
...
+  * @param value the header field value
+  */
+  setExtraHeader: function setExtraHeader(name, value) {
+    this.extraHeaders.push({name: name.toLowerCase(), value: value});
   },
 
-  _sendQueriesBatch: function(params, callback) {
-    return this._jsonRequest({ cache: this.cache,
+  _sendQueriesBatch: function _sendQueriesBatch(params, callback) {
+    function prepareParams() {
```
bobylito: Isn't there a missing return here?

src/AlgoliaSearch.js
```diff
@@ -709,14 +717,13 @@ AlgoliaSearch.prototype = {
...
         // compute the status of the response,
-        var status =
-          // When in browser mode, using XDR or JSONP, we have no statusCode available
-          // So we rely on our API response `status` property.
-          // But `waitTask` can set a `status` property which is not the statusCode (it's the task status)
-          // So we check if there's a `message` along `status` and it means it's an error
-          //
-          // That's the only case where we have a response.status that's not the http statusCode
-          httpResponse && httpResponse.body && httpResponse.body.message && httpResponse.body.status ||
+        var status = // When in browser mode, using XDR or JSONP, we have no statusCode available
```
bobylito: I would comment before the actual not within.

src/browser/migration-layer/is-using-latest.js
```diff
@@ -11,8 +11,8 @@ function isUsingLatest(buildName) {
   var scripts = document.getElementsByTagName('script');
   var found = false;
   for (var currentScript = 0, nbScripts = scripts.length;
-        currentScript < nbScripts;
-        currentScript++) {
+    currentScript < nbScripts;
```
bobylito: Don't you think it's a bit confusing to have the for parts with the same indentation as the inner code?

src/AlgoliaSearch.js
bobylito: I was not aware that some browsers would do such thing :) And it's already a warning, good for me (I haven't seen it in the helper though).

vvo:
> Let me have a look :)

@bobylito I solved all the issues we talked about, and added 'strict mode'. You convinced me :).

jsfmt will always append 'strict mode' on top of every file

bobylito:
Coool :) :+1: 

Becareful seems like we still have issues with Travis.

vvo:
Yes fixing this

PR #126, opened 2015-07-23
fix: withCredentials forced to false in Angular

pixelastic:
This forces the Angular calls made to our API with `withCredentials` set to false (even if the user set them to true on their side, like when behind a Basic Auth)


vvo:
@pixelastic do `npm run lint`

pixelastic:
I've fixed the lint issues. I'll add a `jsfmt` syntax checker to syntastic also

vvo:
pAS

vvo:
@pixelastic you can now do a release! :)

PR #139, opened 2015-09-16
fix: always send a unique response to the user

vvo:


test/spec/common/request-strategy/clone-cached-results.js
```diff
@@ -20,9 +20,11 @@ if (process.browser) {
     });
 
     index.search('something', function(_, firstContent) {
+      firstContent.AHAH = true;
```
pixelastic: Maybe add a comment on why we do this test, or naming `AHAH` something more explicit

(For readers, we do not want objects returned from the cache to be returned by reference, we want real new objects every time)

PR #140, opened 2015-09-16
fix: use version.js instead of version.json

vvo:


vvo:
cc @redox this is the `[Object object]` bug when using some build tools like webpack

PR #153, opened 2015-10-23
feat: new generateSecuredApiKey(apiKey, queryParameters)

vvo:


Nagriar:
@vvo You can also convert the old tagFilters to the new version:
- if it contains a '=' it's already a url encoded query parameters => new version
- if it's an array or a string => convert to { 'tagFilters': value} => new version

vvo:
Yes indeed I forgot that you could also call generateSecuredApiKey(key, tagFiltersAsString), is that what you meant?

vvo:
@Nagriar Not sure to get why we would want to convert every usage to the new key format, won't this introduce some errors because people update the client and still set the x-algolia-query-params?

vvo:
@Nagriar this is ready and doc has been updated, can you have a last look

Nagriar:
No if you use the new format, the x-algolia-query-params is ignored. The new format is simpler to use and to debug. I'd prefer to convert the old tagFilters to the new format too.

vvo:
understood, will change this

vvo:
updated the PR

vvo:
@Nagriar, updated, with unit and functional tests

Nagriar:
great job @vvo!

vvo:
Let's merge this baby

PR #158, opened 2015-10-27
Possible issue with local cache

groovenectar:
We found that the local storage cache only works with `cache[cacheID] = httpResponse.body;`. Is there a reason for it to be `httpResponse.responseText`? We found that to always be null, at least in our case...

### DIFF
--- a/src/AlgoliaSearch.js
+++ b/src/AlgoliaSearch.js
@@ -784,7 +784,7 @@ AlgoliaSearch.prototype = {
         var retry = !ok && Math.floor(status / 100) !== 4 && Math.floor(status / 100) !== 1;
 
         if (client._useCache && ok && cache) {
-          cache[cacheID] = httpResponse.responseText;
+          cache[cacheID] = httpResponse.body;
         }
 
         if (ok) {


vvo:
Hi thanks for the contribution, can I ask you if you are using maybe ou angularjs or jquery client? If so, what version of angular and jquery?

vvo:
I would like to know why the responseText is null in your case, thanks

groovenectar:
We're using both the vanilla JavaScript and AngularJS. We're using version 3.7.8 of "algoliasearch" in bower.json to load the libraries. Not sure why `httpResponse.responseText` is null at this point, but it appeared that `httpResponse.body` contained an object with the results, and caching that instead of `httpResponse.responseText` worked...

vvo:
And thanks I reduced the bug and will submit a fix for this. The thing is that angular.js $http (and I believe you were using the angular js build of our lib) has no "responseText".

vvo:
was released as 3.9.1, thanks for the bug report

groovenectar:
Nice, no problem!

PR #163, opened 2015-11-04
Provides a react native build

bobylito:


redox:
Looks great @bobylito. Did you build a sample app as well? Would be cool to link to an example from the README.

bobylito:
I have a sample project working as well but it definitely need some polish because it is a bit raw right now. I will push it on another repository if you're ok with that.

vvo:
Well looks awesome! I wonder how could we test that. For now it's "beta" I guess and code is mostly the browser build without JSONP etc..

vvo:
@bobylito Feel free to release this

vvo:
> I will push it on another repository if you're ok with that.

Would be cool so that we can point people to it on demand

PR #176, opened 2015-12-04
fix(process.env): set process back to normal in node v4

vvo:


bobylito:
:+1: :dancer:

vvo:
tests ✓

PR #181, opened 2015-12-14
fix(request strategy): always use XHR first then switch to fallback

vvo:


src/AlgoliaSearch.js
```diff
@@ -550,12 +553,27 @@ AlgoliaSearch.prototype = {
       })
     };
 
+    var JSONPParams = map(postObj.requests, function prepareJSONPParams(request, requestId) {
+      return requestId + '=' +
+        encodeURIComponent(
+          '/1/indexes/' + request.indexName + '?' +
```
ElPicador: What happens if request.indexName == `toto?tata=1` ?

test/spec/browser/request-strategy/only-JSONP-when-XHR-failure.js
```diff
@@ -1,63 +0,0 @@
-'use strict';
```
ElPicador: Couldn't you keep the test but change it for the new behavior?

vvo: Indeed, so a test that would first use XHR (fail), then JSONP and then use XHR again?

src/AlgoliaSearch.js
vvo: So you mean encoding the indexName inside the already encoded param?

test/spec/browser/request-strategy/only-JSONP-when-XHR-failure.js
ElPicador: Yeah, because it seems a tricky behavior so a test should help it debug/no break it

src/AlgoliaSearch.js
ElPicador: Well, I understand that `encodeURIComponent` encodes the URI, but not oh does it handle encoding `'/1/indexes/toto?tata=1?queryString=tutu'`. 
I'm just curious about the sides effects

vvo: It is true that in most methods we do: `/1/indexes/ + encodeURIComponent(indexName)` so I guess we should do the same here.

Then how it's handled really is a question for the API server, I do not have more info here

ElPicador: Ok, thanks for the answer :smile:

test/spec/browser/request-strategy/only-JSONP-when-XHR-failure.js
vvo: fixed in 9a56fa9113d093a16710452e981acfbf8bc373da

src/AlgoliaSearch.js
vvo: fixed in 9a56fa9113d093a16710452e981acfbf8bc373da

ElPicador:
LGTM

vvo:
THX @ElPicador for review :)

ElPicador:
You are most welcome :smiley:

PR #186, opened 2015-12-28
Allow specification of cache

dustincoates:
This allows for offline use of the search helper (e.g. in development situations) as a cache can be recorded before-hand and used if no connection is detected.

### DIFF
--- a/test/spec/common/client/set-cache.js
+++ b/test/spec/common/client/set-cache.js
@@ -1,0 +1,49 @@
+'use strict';
+
+var test = require('tape');
+
+test('providing a cache when initializing', function(t) {
+  t.plan(1);
+
+  var fauxJax = require('faux-jax');
+  var getCredentials = require('../../../utils/get-credentials');
+  var credentials = getCredentials();
+
+  fauxJax.install();
+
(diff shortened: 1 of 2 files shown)


vvo:
Cool, can you add a comment stating that in some situations you might want to warm the cache?

And add a test too so that we do not break this feature?

vvo:
cc @dustincoates

dustincoates:
@vvo Indeed. Will do both. :+1:

dustincoates:
I'll take a look at that failing test. Doesn't seem related to what I added, so may take a little bit of time to pin down.

vvo:
careful you commited the builded files, can you revert them?

test/spec/common/client/set-cache.js
```diff
@@ -0,0 +1,22 @@
...
+    cache: {
+      'path/for/query': {}
+    }
+  };
+
+  var algoliasearch = require('../../../../');
+  var search = algoliasearch('applicationID', 'apiKey', opts);
+
+  t.deepEqual(
+    search.cache,
```
vvo: The best way to test this is to actually initialize a client with some fake app id and index then check that when doing a query you will get the cached response you provided, can you do so?

test/spec/common/client/set-cache.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+var test = require('tape');
+
+test('cache is not empty object when specified in opts', function(t) {
```
vvo: test('providing a cache when initializing')

test/spec/common/client/set-cache.js
vvo: This will decouple the test from the internal implementation that might not store the cache inside search.cache in the future, but maybe in another property/hidden.

dustincoates:
@vvo Sorry, thought I was supposed to build before committing. Will revert and make those changes.

dustincoates:
@vvo Let me know if this is in line with what you had in mind.

test/spec/common/client/set-cache.js
```diff
@@ -0,0 +1,48 @@
...
+    cache: cache
+  };
+
+  var algoliasearch = require('../../../../');
+  var searchWithCache = algoliasearch(credentials.applicationID, credentials.searchOnlyAPIKey, opts);
+
+  var indexWithCache = searchWithCache.initIndex(credentials.indexName);
+
+  fauxJax.on('request', function(req) {
+    req.respond(200, {}, cachedResult);
```
vvo: You should not even have to use fauxJax, since if there's a request it means that your "prefill" cache feature failed.

What I would do is put a t.fail() inside the .on('request') instead.

vvo:
> @vvo Let me know if this is in line with what you had in mind.

We're almost good, let me know, sorry for late reply

test/spec/common/client/set-cache.js
dustincoates: Oh yeah, great point. :+1:

vvo: Let me know @dustincoates when you have finished this PR (t.fail?)

dustincoates: @vvo I'm working on this today or tomorrow at the very latest.

vvo:
rebased into a single commit

PR #189, opened 2016-01-08
If brew and nvm then use it

ElPicador:
The idea is to be able to use `nvm` installed with `brew`


vvo:
npm test was not working for you because nvm was installed with brew?

ElPicador:
Exactly, I don't have `~/.nvm/nvm.sh` so the script `scripts/test-node` failed

Issue #195, opened 2016-02-02
V4 Roadmap


vvo:
Here are some V4 goals I would like to tackle:
- [ ] switch codebase to ES6
- [ ] allow building modular clients (search only, places). Mostly by extracting the requester into a single object
- [ ] switch to webpack
- [ ] automatic releases through npm and jsdelivr
- [ ] monorepo architecture with packages/algoliasearch-[jquery|angular..]
- [ ] publish packages/ so that they can be required easily
- [ ] experiment with rollup (seems to be the promising ES6 focused module builder, low overhead)
- [ ] automatic changelog
- [ ] drop non-CORS browsers
- [ ] drop non ES5 browsers
- [ ] provide a search only client through a specific package
- [ ] provide a specific places client through a specific package
- [ ] stop building complex stateful prototype based objects, use simple wrappers
- [ ] do real unit testing, drop the testing stack
- [ ] use mocha, jsdom
- [ ] default to https for write actions, browser protocol for reads
- [ ] review API #204 
- [ ] remove all legacy signatures and already deprecated methods in V3
- [ ] provide a compatibility layer package to ease the migration, with clear messages on what to change
- [ ] generateSecuredAPIKey should be static (maybe other methods)
- [ ] clearly identify public/private props
- [ ] requests should be cancelable
- [ ] drop callback support, promises only, show how to still use callbacks
- [ ] bootstrap website
- [ ] full jsdoc (see https://esdoc.org/) on website
- [ ] single cache, no more client AND index cache
- [ ] run unit tests with karma (maybe)
- [ ] code coverage and automation
- [ ] use greenkeeper.io
- [ ] monitor V3 /latest usage before publishing
- [ ] drop jsdelivr /latest compatibility layer (we automatically load V2 when used with /latest/ on jsdelivr)
- [ ] drop shrinkwrap
- [x] introduce read and write timeouts
- [ ] library should work in service workers
- [ ] client instantiation should be an option object even for apiKey and appId, this is already the case of instantsearch.js for example. Or docsearch
- [ ] add a .preconnect method on client that will initiate the connection (use case: click in input)

kievechua:
Hi @vvo, when can we expect generateSecuredApiKey implemented?
This issue been for more than a year.
Also didn't see any rest api I can implement it.
Can you point me to correct direction?

Haroenv:
You should be able to do `client.generateSecuredApiKey()` on your instance already @kievechua

[...]

PR #196, opened 2016-02-05
feat(falback): provide opts.useFallback to avoid using JSONP

vvo:


src/AlgoliaSearch.js
```diff
@@ -103,6 +103,7 @@ function AlgoliaSearch(applicationID, apiKey, opts) {
 
   this._ua = opts._ua;
   this._useCache = opts._useCache === undefined || opts._cache ? true : opts._useCache;
+  this._useFallback = opts.useFallback === undefined ? true : opts.useFallback;
```
ElPicador: can't you just do a `this._useFallback = !!opts.useFallback` ?

vvo: I try to avoid "smart" notations where you have to double think what's it's doing to the provided value

src/AlgoliaSearch.js
```diff
@@ -685,7 +686,16 @@ AlgoliaSearch.prototype = {
     var client = this;
     var tries = 0;
     var usingFallback = false;
-    var hasFallback = client._request.fallback && initialOpts.fallback;
+    var hasFallback = client._useFallback && client._request.fallback && initialOpts.fallback;
+    var headers;
+
+    if (this.apiKey.length > 500 && initialOpts.body !== undefined && initialOpts.body.params !== undefined) {
```
ElPicador: I would have put the `500` in a constant somewhere (magic number)

src/AlgoliaSearch.js
ElPicador: Ok seems legit :smile:

vvo:
There's clearly no tests for this, as part of #195 I want to stop doing tests like I designed before (black box) and do unit testing + separation of concerns. Thus the jsonRequest should be an independent object not directly sharing properties with the client if feasible

src/AlgoliaSearch.js
vvo: yes, adding

vvo:
The way tests are written right now would require us to test every method using POST and sending it with a very long API key, not so interesting/nice test design :)

ElPicador:
I understand for the tests, so LGTM

src/AlgoliaSearch.js
vvo: Same than I avoid:

```js
var int = +string;

if (~someStr.indexOf("a")) {
	// Found it
} else  {
	// Not Found
}

// and many other notations I find hard to read for beginners, and me
```

PR #198, opened 2016-02-14
Readme cleanup

asurak:
When using the readme, I've hit a couple of issues that create extra confusion or unnecessary overhead. So I've:
- removed extra redirections
- added HTTPS by default on all www.algolia.com
 - since www.algolia.com has HSTS, the browser will do it automatically but if it a first request, there is unnecessary delay in reaching the documentation caused by the redirection to HTTPS
- fixed non-existent anchors to the documentation
 - most of the used anchors don't exist in the documentation anymore and the URLs had old format

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 
 
-[Algolia Search](http://www.algolia.com) is a hosted full-text, numerical, and faceted search engine capable of delivering realtime results from the first keystroke.
+[Algolia Search](https://www.algolia.com) is a hosted full-text, numerical, and faceted search engine capable of delivering realtime results from the first keystroke.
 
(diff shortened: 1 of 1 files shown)


redox:
Keep in mind we need to replicate the changes here https://github.com/algolia/algoliasearch-client-readme-generator as well.

asurak:
Hmm, seems like I have merged a wrong repo. I have the change ready even for the parent repo ;)

asurak:
Here is the PR in the parent repo: https://github.com/algolia/algoliasearch-client-readme-generator/pull/18

vvo:
So we close this one?

PR #211, opened 2016-03-29
Updated README (prepare algolia.com's integration)

redox:
cc @alexandremeunier

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,18 @@
+<!--NO_HTML-->
+
 # Algolia Search API Client for JavaScript
 
+<!--/NO_HTML-->
 
 
 
+
+
+<!--NO_HTML-->
 
 [Algolia Search](https://www.algolia.com) is a hosted full-text, numerical, and faceted search engine capable of delivering realtime results from the first keystroke.
+
+<!--/NO_HTML-->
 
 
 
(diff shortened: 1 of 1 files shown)


vvo:
woa, shiny. Will we be able to have a list of methods at some point on top of this? Rather than use cases list. I guess once you are here what you are looking for is a method list. Dunno if that's doable given all the various API clients still

vvo:
But it's already a looooooooot better than previously :)

alexandremeunier:
@vvo: 100% agree. We need a more explicit method list. That's my next step :)

redox:
@vvo yeah, the method list is supposed to be  
screen shot 2016-03-29 at 15 01 26


but indeed, it's not code.

redox:
Not sure why the tests are failing :/

```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

Saucelabs issue?

vvo:
Yeah saucelabs issue.

@redox what I meant by method list is this:

- index.setSettings
- index.getSettings
- index.search
- client.search
- index.getObject
- client.batch
- ...

So that once you already know the client (and if you are on github, there's a high probability you already know it) then you do not have to map "Ah so I am looking for index.getSettings usafe, let's find where it is in this list".

I agree this may no be done easily with what we have. But this is ultimately what I am looking for when I am on our API client github projects: the API list, not use cases.

This is similar to instantsearch.js where in the menu we have "searchBox" versus "add a search input"

vvo:
This can be merged without any issue

vvo:
LGTM

PR #217, opened 2016-04-22
feat(build): provide a lite build

### LINKED ISSUE #215: Client is heavyweight
Hi, I noticed that adding the client to my application added 70kb to the minified build. Is that normal?
Could we split the client to have only the minimal code on the client?

vvo:
Now providing a lite browser build:
- search only
- vanilla JS only (no angular, jquery lite builds for now)
- same export name (`algoliasearch`)
- now 5.93kb gzipped (versus 20kb)
- IE11/10/9 not natively supported, use http://cdn.polyfill.io/v2/docs/
  or just provide the Promise polyfill in the global object (window)
- same init usage
- has
  client.search/index.search/clearCache/setRequestTimeout/setExtraHeader/..
- no write/indexing/waitTask/settings methods => search only
- available on jsDelivr using algoliasearchLite[.min].js
- also requireable using
  var algoliasearch = require('algoliasearch/dist/algoliasearchLite.js');

fixes #215


redox:
That's awesome @vvo!

vvo:
5.93kb

:dancers: 

We can have IE9/10/11 compat but it would weight something like 7-8kb, what do you think?

vvo:
It's 8KB with the Promise polyfill. I will not do it for now, let's wait for some users to ask for it before making plans

anaibol:
Personally I prefer without polyfills. ES6 modules make it easy to add extra polyfills if required.

vvo:
So this is passing tests, lets wait monday morning for a release :)

anaibol:
By the way, how can we "ES6 import" this lite version? Maybe it should be a separate NPM package.

vvo:
`import 'algoliasearch' from 'algoliasearch/dist/algoliasearchLite.js'` should do it.. once released.

In the meantime you should be able to build it yourself if you want:

```
npm install
NODE_ENV=production npm run build
```

anaibol:
Great!

redox:
> We can have IE9/10/11 compat but it would weight something like 7-8kb, what do you think?

Are we able to add the Promise polyfill only for the `<script>` build? If so I would go that way!

PR #221, opened 2016-05-01
feat(synonyms): add new synonyms API

vvo:
Tests will succeed once synonyms API is deployed everywhere.


Nagriar:
@vvo We can merge it, the API is deployed

vvo:
WOOHOO

src/Index.js
```diff
@@ -652,12 +652,109 @@ Index.prototype.getSettings = function(callback) {
   var indexObj = this;
   return this.as._jsonRequest({
     method: 'GET',
-    url: '/1/indexes/' + encodeURIComponent(indexObj.indexName) + '/settings',
+    url: '/1/indexes/' + encodeURIComponent(indexObj.indexName) + '/settings?getVersion=2',
```
olance: This might "break" some implementations. Any client that uses getSettings to get synonyms in some sync scripts won't have access to them anymore, so they might trigger unnecessary index build by sending synonyms again.

I think by default `getVersion` should be omitted and there should be an option to set it. Something like `getSettings({ version: 2 })`?

src/Index.js
```diff
@@ -652,12 +652,109 @@ Index.prototype.getSettings = function(callback) {
...
+    callback = params;
+    params = {};
+  } else if (params === undefined) {
+    params = {};
+  }
+
+  return this.as._jsonRequest({
+    method: 'POST',
+    url: '/1/indexes/' + encodeURIComponent(this.indexName) + '/synonyms/search',
+    body: params,
```
olance: It could be nice to handle the case when `params.type` is an Array.
It would need to be joined as a comma-separated list of values.

src/Index.js
vvo: > Any client that uses getSettings to get synonyms in some sync scripts

Can you explain the issue a bit more? Every method call is already asynchronous by default, then if getVersion=2 triggers a different behavior than before, I was not aware of that.

src/Index.js
vvo: The API should handle all array properties being either a JSON encoded array or a comma separated list of values. Did you experience any issue?

src/Index.js
vvo: Ok just saw the note about getVersion=2 will no more return synonyms. I will update the code accordingly.

vvo: But.. Seems like other API clients have done the same somehow: https://github.com/algolia/algoliasearch-client-php/blob/master/src/AlgoliaSearch/Index.php#L693

vvo: Same for ruby, https://github.com/algolia/algoliasearch-client-ruby/blob/93f7a87a839f43a5d7e06bd9385a837083e34468/lib/algolia/index.rb#L436

Not saying this is good, just searching

redox: Yes forcing synonyms version 2 is a choice we've done. I don't think we want to expose the complexity of such versioning to our users. I think we need to live with it like that.

olance: Well @redox we do have to expose it some way or another, because I don't think we're the kind to ship something not backward compatible and not warn about it?

So we'll at least have to include a warning in all readmes to tell people they have to use `searchSynonyms` because now the settings don't include synonyms anymore.

src/Index.js
olance: Yes, the API rejects the array with the following error:
`{"message":"pbject cannot contains array","status":400}`

vvo: Hm maybe we should support that then because the API supports comma separated and array most of the time.

olance: @Nagriar / @speedblue, care to comment? :)

PR #254, opened 2016-05-20
Fix #251: increments hostIndex (host adress) on fallback

### LINKED ISSUE #251: DNS fallback resolution not working correctly
Following the recent NSOne DNS attack, we had a few customers reaching to use because their search wasn't working. For most of them it was because they were using an old API client version without the DNS fallback mechanism.

But for a few of them, that were using the very latest (3.14) version, they still had issues where some requests were targetting `algolia.net` and others `algolianet.com`. One user confirmed that reverting to 3.13 fixed the issue, the other monkey patched the code to force the `algolianet.com` endpoint.

I haven't been able to reproduce it on my side (DNS correctly working for me), but we did change this code in between those two releases, so we might have introduced a bug. I will dive more deeply into it as soon as I can but I'm taking a 11h flight tomorrow morning so not sure I'll be able to really focus on it.

All help is thus welcome into finding the bug and fix it as fast as we can.

bobylito:
This issue was caused because we do not distinguish CORS errors from DNS
resolution one and the code for incrementing the host index was moved into
the part that switches to the fallback.

Truth is that the current solution is not optimal as we are switching to
JSONP when we can't resolve the server name.


bobylito:
Explanation note:
 - before / working: https://github.com/algolia/algoliasearch-client-js/blob/3.13.1/src/AlgoliaSearch.js#L864-L871
 - current state on master/broken: https://github.com/algolia/algoliasearch-client-js/blob/master/src/AlgoliaSearchCore.js#L363-L369

pixelastic:
Seems to fix the issue, I'll work on adding tests to cover it.

We'll also discovered that our jsonp fallback strategy is interfering with the DNS fallback. Apparently, there is no way to tell if the error is a DNS one or a CORS one. We'll need to go deeper into that. But first step will be to release a 3.15 that fixes the DNS issue

pixelastic:
I did not manage to add the missing test for that so far...

bobylito:
Closing pull request as it is not targeted towards the good branch...

PR #256, opened 2016-05-22
Fix #251: increments hostIndex (host adress) on fallback

### LINKED ISSUE #251: DNS fallback resolution not working correctly

bobylito:
NB: See #254 for previous discussions

This issue was caused because we do not distinguish CORS errors from DNS
resolution one and the code for incrementing the host index was moved into
the part that switches to the fallback.

Truth is that the current solution is not optimal as we are switching to
JSONP when we can't resolve the server name.

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -360,19 +360,20 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
         return client._promise.reject(err);
       }
 
-      if (err instanceof errors.RequestTimeout) {
+      if (err instanceof errors.RequestTimeout || err instanceof errors.NodeNetwork) {
         return retryRequest();
       } else if (!usingFallback) {
-        client.hostIndex[initialOpts.hostType] = ++client.hostIndex[initialOpts.hostType] % client.hosts[initialOpts.hostType].length;
         // next request loop, force using fallback for this request
         tries = Infinity;
       }
+
+      client.hostIndex[initialOpts.hostType] = (client.hostIndex[initialOpts.hostType] + 1) % client.hosts[initialOpts.hostType].length;
 
       return doRequest(requester, reqOpts);
     }
 
     function retryRequest() {
-      client.hostIndex[initialOpts.hostType] = ++client.hostIndex[initialOpts.hostType] % client.hosts[initialOpts.hostType].length;
+      client.hostIndex[initialOpts.hostType] = (client.hostIndex[initialOpts.hostType] + 1) % client.hosts[initialOpts.hostType].length;
       reqOpts.timeout = client.requestTimeout * (tries + 1);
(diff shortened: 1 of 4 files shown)


vvo:
Cleaned dist/ and removed keys, updated commit message with:

```
When we are switching to fallback, increment host index so that JSONP
will use another host.

This issue was caused because we cannot distinguish CORS errors from DNS
resolution errors in the browser.

And the code for incrementing the host index was moved into the part
that switches to the fallback in
4ec5e6a.

Truth is that the current solution is not optimal as, in the browser,
we are switching to JSONP when we can't resolve the server name.

This is only done because we are not sure how CORS request can be
blocked (synchronously, asynchronously).

In a next version we would drop it and wait for real issues to show up.
74f1d54  
```

src/AlgoliaSearchCore.js
```diff
@@ -360,19 +360,20 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
...
       }
 
+      client.hostIndex[initialOpts.hostType] = (client.hostIndex[initialOpts.hostType] + 1) % client.hosts[initialOpts.hostType].length;
+
       return doRequest(requester, reqOpts);
     }
 
     function retryRequest() {
-      client.hostIndex[initialOpts.hostType] = ++client.hostIndex[initialOpts.hostType] % client.hosts[initialOpts.hostType].length;
+      client.hostIndex[initialOpts.hostType] = (client.hostIndex[initialOpts.hostType] + 1) % client.hosts[initialOpts.hostType].length;
```
redox: From what I understand, not incrementing that `hostIndex` here will result in a loop always using the same host in case of `RequestTimeout` or `NodeNetwork`; is that correct?

bobylito: Yes, you are right @redox

redox: So that's a issue, right? In case of timeout/error/whatever: we always want to retry on the next host.

redox: Oh my bad, didn't get the assignment `client.hostIndex[initialOpts.hostType] =` -> we **ARE** incrmenting the `hostIndex` 👍

PR #258, opened 2016-05-23
docs(cache): Add a more verbose explanation of the cache mechanism

pixelastic:
I realized that the previous `cache` documentation was a bit succinct,
so I just expanded it with more context, explanation and examples.


pixelastic:
Closing. Should have done it through the README generator.

PR #259, opened 2016-05-24
docs(example): Replace autocomplete/instantsearch example with gif

pixelastic:
I have submitted another PR on the readme generator (https://github.com/algolia/algoliasearch-client-readme-generator/pull/59) to give more emphasis on the autocomplete.js and instantsearch.js libraries. The goal is to make sure most people are aware of those libs and use them instead of building their UI from scratch.

With that goal in mind, I removed the autocomplete and instantsearch example and replaced them with animated gif. The readme will be updated (in the other PR) with a full explanation of what those libs are doing.


bobylito:
@pixelastic you removed examples that are useful for either testing (part of our release involves testing the build against the examples) and for learning the lib (as a developper I like having samples that uses the library). I don't think we should remove them.

vvo:
Yes @pixelastic the examples are still useful when developing, having both seems better. GIFS for ease of reading and code for hacking

pixelastic:
Arf, I wasn't aware of that, my bad. I'll add them back, maybe with a comment pointing to autocomplete.js / instantsearch.js.

pixelastic:
I've re-added the HTML files, adding a small comment inside redirecting to autocomplete/instantsearch.

pixelastic:
image

Here is the error from Travis. Not sure how this is relevant. I'm lost at debugging Travis

vvo:
LGTM.

The current testing stack is flacky because of the underlying tunneling tool we are using to allow browsers from saucelab to access travis started web servers serving tests. It's a mess. I test locally for now

pixelastic:
When you say that you test locally, do you run a saucelab local test, or manual testing on the browsers your have available on your machine?

vvo:
@pixelastic I use npm run dev then launch virtualbox and access my 192.168.56.1 (host) to launch the test.

I have old browsers running with https://github.com/xdissent/ievms/

Also, I just switched back to "ngrok" (it was buggy also for some time) for our tunneling between saucelabs and travis and it's now much more stable in PRS.

PR #262, opened 2016-05-24
fix(retry): also retry on non search methods when DNS failure

### LINKED ISSUE #250: JSONP fallback is too aggressive & wrong when angularJS
In the following code, we assign the `tries` variable to `Infinity` to force the JSONP retry. But that's wrong:
 - when using the Angular build, `hasFallback` is false (because `client._useFallback` is undefined) and we're therefore not retrying
 - when there is any network error on 1 host, we need to retry on the others without going for JSONP

```js
      // stop the request implementation when:
      if (
        // we did not generate this error,
        // it comes from a throw in some other piece of code
        err instanceof errors.Unknown ||

        // server sent unparsable JSON
        err instanceof errors.UnparsableJSON ||

        // max tries and already using fallback or no fallback
        tries >= client.hosts[initialOpts.hostType].length &&
        (usingFallback || !hasFallback)) {
        // stop request implementation for this command
        err.debugData = debugData;
        return client._promise.reject(err);
      }

      if (err instanceof errors.RequestTimeout) {
        return retryRequest();
      } else if (!usingFallback) {
        client.hostIndex[initialOpts.hostType] = ++client.hostIndex[initialOpts.hostType] % client.hosts[initialOpts.hostType].length;
        // next request loop, force using fallback for this request
        tries = Infinity;
      }

      return doRequest(requester, reqOpts);
    }
```

vvo:
Before this commit, methods with no fallback support (basically every
method but search) would fail at retry if the DNS error occured before the API
client timeout.

The behavior is now:
- when a method with a fallback errors because of DNS failure we will
switch to JSONP right away
- when a method with no fallback errors because of DNS failure we will
still retry

We could use the same mechanism for both (= always try all hosts
before JSONP) but I am not confident doing this change within a patch
or minor version. We have too litle data on how blocked XHRS are
triggered (async, sync?).

fixes #250

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -273,16 +273,8 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
       requestDebug('received response: statusCode: %s, computed statusCode: %d, headers: %j',
         httpResponse.statusCode, status, httpResponse.headers);
 
-      var ok = status === 200 || status === 201;
-      var retry = !ok && Math.floor(status / 100) !== 4 && Math.floor(status / 100) !== 1;
-
-      if (client._useCache && ok && cache) {
-        cache[cacheID] = httpResponse.responseText;
-      }
-
-      if (ok) {
-        return httpResponse.body;
-      }
+      var httpResponseOk = Math.floor(status / 100) === 2;
+      var shouldRetry = Math.floor(status / 100) !== 4 && Math.floor(status / 100) !== 2;
 
       var endTime = new Date();
       debugData.push({
@@ -298,11 +290,22 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
         duration: endTime - startTime
       });
 
-      if (retry) {
+      if (httpResponseOk) {
+        if (client._useCache && cache) {
+          cache[cacheID] = httpResponse.responseText;
+        }
+
+        return httpResponse.body;
+      }
+
+      if (shouldRetry) {
         tries += 1;
         return retryRequest();
       }
 
+      requestDebug('unrecoverable error');
+
+      // no success and no retry => fail
       var unrecoverableError = new errors.AlgoliaSearchError(
         httpResponse.body && httpResponse.body.message, {debugData: debugData}
       );
@@ -360,27 +363,22 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
         return client._promise.reject(err);
       }
 
-      // When a timeout occured or if a network error occured and we have no fallback
-      if (err instanceof errors.RequestTimeout || !hasFallback) {
-        return retryRequest();
+      // When a timeout occured, retry by raising timeout
+      if (err instanceof errors.RequestTimeout) {
+        return retryRequestWithHigherTimeout();
       }
 
-      // The only case where we reach here is a Network error occured, there's a fallback (JSONP), we use it
-      // we cannot easily distinguish blocked XHRS from DNS request failure in a cross browser way.
-      // So we use the fallback as soon as it's available and there was a network error, be it DNS failure or
-      // blocked XHR.
-
-      if (!usingFallback) {
-        // next request loop, force using fallback for this request
-        tries = Infinity;
-      }
-
-      client.hostIndex[initialOpts.hostType] = (client.hostIndex[initialOpts.hostType] + 1) % client.hosts[initialOpts.hostType].length;
-
-      return doRequest(requester, reqOpts);
+      return retryRequest();
     }
 
     function retryRequest() {
+      requestDebug('retrying request');
+      client.hostIndex[initialOpts.hostType] = (client.hostIndex[initialOpts.hostType] + 1) % client.hosts[initialOpts.hostType].length;
+      return doRequest(requester, reqOpts);
+    }
+
+    function retryRequestWithHigherTimeout() {
+      requestDebug('retrying request with higher timeout');
       client.hostIndex[initialOpts.hostType] = (client.hostIndex[initialOpts.hostType] + 1) % client.hosts[initialOpts.hostType].length;
       reqOpts.timeout = client.requestTimeout * (tries + 1);
       return doRequest(requester, reqOpts);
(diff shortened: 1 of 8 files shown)


vvo:
cc @redox @pixelastic @bobylito

redox:
> We could use the same mechanism for both (= always try all hosts
> before JSONP) but I am not confident doing this change within a patch
> or minor version.

I would still go for it :) I don't see any risk of retrying on different hosts before going for JSONP. What do you fear?

vvo:
> What do you fear?

The unknown! Blocked XHRS may be asynchronously blocked by a proxy and we would wrongly retry and retry untill we tried all hosts before using JSONP. Only assumptions, not based on actual data :dancers:

redox:
> The unknown! Blocked XHRS may be asynchronously blocked by a proxy and we would wrongly retry and retry untill we tried all hosts before using JSONP. Only assumptions, not based on actual data 👯

But in the worst case, we just delay a bit the JSONP retry; right? (until the other CORS requests timeout/fail)

vvo:
> we just delay a bit the JSONP retry; right?

Yes worse case but it would be like a verrrrrrrrrrrry long time instead of 2s right now (timeout is incremental * number of req).

If we do not get any support issue from doing this then I don't even know why do we support JSONP "automatically" (instead of an option to force JSONP) at all. If no one notice that instead of a 2s delay for searches they get 30s then no one is using JSONP fallback or monitoring users using it.

pixelastic:
My gut feeling is that no one is using JSONP anymore :/

test/run-integration.js
```diff
@@ -330,13 +331,33 @@ function dnsFailThenSuccess(t) {
...
 
+function dnsFailThenSuccessNoSearch(t) {
+  t.plan(1);
+
+  var client_ = algoliasearch(
+    appId,
+    apiKey, {
+      // .biz is a black hole DNS name (not resolving)
+      hosts: [appId + '-dsn.algolia.biz', appId + '-dsn.algolia.net'],
+      timeout: 120000, // let's wait for the DNS timeout
```
odeke-em: A minor one, could we constify this timeout e.g
```javascript
const dnsTimeout = 120000;
```
and then use it in the 3 places that `120000` is being used? Am asking for this as I noticed that the value `40000` was changed to `120000` suggesting that the timeout was realized to previously be short.

src/AlgoliaSearchCore.js
```diff
@@ -360,9 +360,17 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
...
       }
 
-      if (err instanceof errors.RequestTimeout || err instanceof errors.NodeNetwork) {
+      // When a timeout occured or if a network error occured and we have no fallback
+      if (err instanceof errors.RequestTimeout || !hasFallback) {
         return retryRequest();
-      } else if (!usingFallback) {
+      }
+
+      // The only case where we reach here is a Network error occured, there's a fallback (JSONP), we use it
```
odeke-em: I don't quite follow the comment above. Did you mean to say?
```diff
// The only case where we reach here is a Network error occured,
-, there's a fallback (JSONP), we use it
+ and if there is a fallback (JSONP), we'll use it 
```

or 
```diff
-, there's a fallback (JSONP), we use it
+ and there's a fallback (JSONP) that we can use
```

vvo: I will clarify this comment

vvo:
Let's not optimize JSONP switch with tries = unlimited

src/server/builds/node.js
```diff
@@ -210,7 +210,7 @@ AlgoliaSearchNodeJS.prototype._request = function request(rawUrl, opts) {
 
       abort();
       clearTimeout(timeoutId);
-      reject(new errors.NodeNetwork(err.message, err));
+      reject(new errors.Network(err.message, err));
```
bobylito: Isn't Network the error sent by the client when there is a CORS error? It might be a bit misleading if we have errors that are bags for several symptoms...

src/AlgoliaSearchCore.js
```diff
@@ -360,9 +360,17 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
...
+      // we cannot easily distinguish blocked XHRS from DNS request failure in a cross browser way.
+      // So we use the fallback as soon as it's available and there was a network error, be it DNS failure or
+      // blocked XHR.
+
+      if (!usingFallback) {
```
bobylito: You should probably test if there is indeed a fallback?

test/run-integration.js
bobylito: 👍 even though this is not ES6. A good old var with caps can do the trick, right?

vvo: will add

src/server/builds/node.js
vvo: Yes it is. I just made a new commit where we always try all hosts when there's a network error before switching to JSONP.

Trying to switch to JSONP early was made by thinking the only network error that could happen was blocked XHR. While it can be a DNS failure on the browser, we should retry using XHR.

src/AlgoliaSearchCore.js
```diff
@@ -298,11 +290,22 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
         duration: endTime - startTime
       });
 
+      if (ok) {
```
bobylito: ok? Can you name it something like `httpResponseOk`?

src/server/builds/node.js
bobylito: Cool :)

src/AlgoliaSearchCore.js
vvo: Yes will do

test/spec/browser/request-strategy/no-JSONP-when-4xx-or-1xx.js
```diff
@@ -4,7 +4,6 @@ var test = require('tape');
 
 test('Request strategy does not fallback to JSONP', function(t) {
   noJSONP(t, 404);
-  noJSONP(t, 101);
```
vvo: This is not in the request strategy spec and thus was removed

test/spec/browser/request-strategy/no-JSONP-when-4xx.js
```diff
@@ -4,7 +4,6 @@ var test = require('tape');
 
 test('Request strategy does not fallback to JSONP', function(t) {
   noJSONP(t, 404);
-  noJSONP(t, 101);
```
vvo: This is not in the request strategy spec and thus was removed

PR #275, opened 2016-06-03
Made compatible with Angular 1.4.3

wdollar:


vvo:
This seems very nice @wdollar! Just for the record, could you point us to some documentation/readings as to why our usage was not good with angular 1.4.3?

Could we make the JavaScript angular client somehow better to handle that case more easily with angular?

PR #320, opened 2016-08-30
fix(client.search): accept very long API keys

vvo:


package.json
```diff
@@ -101,7 +101,7 @@
     "sinon": "^1.17.4",
     "superagent": "^2.0.0",
     "tape": "^4.5.1",
-    "uglify-js": "^2.7.3",
+    "uglify-js": "2.6.4",
```
bobylito: Why the downgrade?

vvo: Shame we cannot comment in package.json, so it was added to the commit message: `also fix uglify-js version because it's buggy in IE8:
mishoo/UglifyJS2#1039`

vvo: basically uglify-js dropped IE8 support just like that :D

PR #331, opened 2016-09-20
adding typescript definition

CBaptiste:
Hello,

Let's integrate the whole typescript definition of algoliasearch !

### DIFF
--- a/typings/algoliasearch.d.ts
+++ b/typings/algoliasearch.d.ts
@@ -1,0 +1,1565 @@
+// Type definitions for algoliasearch-client-js 3.18.1
+// Project: http://github.com/algolia/algoliasearch-client-js
+
+
+/**
+ * Describes the answer from algolia's api after a search operation
+ */
+interface IAlgoliaResponse {
+    /**
+     * Contains all the hits matching the query
+     * https://github.com/algolia/algoliasearch-client-js#response-format
+     */
+    hits:any[];
+    /**
+     * Current page
+     * https://github.com/algolia/algoliasearch-client-js#response-format
+     */
+    page:number;
+    /**
+     * Number of total hits matching the query
+     * https://github.com/algolia/algoliasearch-client-js#response-format
+     */
+    nbHits:number;
+    /**
+     * Number of pages
+     * https://github.com/algolia/algoliasearch-client-js#response-format
+     */
+    nbPage:number;
+    /**
+     * Number of hits per pages
+     * https://github.com/algolia/algoliasearch-client-js#response-format
+     */
+    hitsPerPage:number;
+    /**
+     * Engine processing time (excluding network transfer)
+     * https://github.com/algolia/algoliasearch-client-js#response-format
+     */
+    processingTimeMS:number;
+    /**
+     * Query used to perform the search
+     * https://github.com/algolia/algoliasearch-client-js#response-format
+     */
+    query:string;
+    /**
+     * GET parameters used to perform the search
+     * https://github.com/algolia/algoliasearch-client-js#response-format
+     */
+    params:string;
+}
+/*
+ Interface for the algolia client object
+ */
+interface IAlgoliaClient {
+    /**
+     * Initialization of the index
+     * @param name: index name
+     * return algolia index object
+     * https://github.com/algolia/algoliasearch-client-js#init-index---initindex
+     */
+    initIndex(name:string):IAlgoliaIndex;
+    /**
+     * Query on multiple index
+     * @param queries index name, query and query parameters
+     * @param cb callback(err, res)
+     * https://github.com/algolia/algoliasearch-client-js#multiple-queries---multiplequeries
+     */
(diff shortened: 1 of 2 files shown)


CBaptiste:
I need to double check all def and properties but I think it is ok.
what do you think @vvo ?

CBaptiste:
👍 
ts

typings/algoliasearch.d.ts
```diff
@@ -0,0 +1,1069 @@
...
+    alternativesAsExact?: any;
+    /**
+     * The name of the attribute used for the Distinct feature
+     * default: null
+     */
+    attributeForDistinct?: string;
+    /**
+     * If set to 1, enables the distinct feature, disabled by default, if the attributeForDistinct index setting is set.
+     */
+    distinct?: number;
```
redox: `distinct` can also be `true` or `false`. Not sure how to handle it :/

redox: Looks like we need to use `any`.

CBaptiste: yes yes, for dynamically typed param, we ll use any !
thanks for spotting it out !

typings/algoliasearch.d.ts
```diff
@@ -0,0 +1,1252 @@
...
+     */
+    initIndex(name:string):IAlgoliaIndex;
+    /**
+     * Initialization of the client
+     * @param name of the index
+     * @param params options related to the client initialization
+     * return algolia index object
+     * https://github.com/algolia/algoliasearch-client-js#init-index---initindex
+     */
+    initIndex(name:string, params:IClientOptions):IAlgoliaIndex;
```
vvo: I think there's no way to call initIndex(name, params), can you link the code?

typings/algoliasearch.d.ts
```diff
@@ -0,0 +1,1252 @@
...
+     */
+    search(queries:{ indexName:string, query:string, options:IAlgoliaQueryParameters }, params:any, cb:Function);
+    /**
+     * Query on multiple index
+     * @param queries index name, query and query parameters
+     * @param params query parameter
+     * @param cb callback(err, res)
+     * https://github.com/algolia/algoliasearch-client-js#multiple-queries---multiplequeries
+     */
+    search(queries:{ indexName:string, query:string, options:IAlgoliaQueryParameters }, params:any, cb:Function, err:Function);
```
vvo: There's no err:Function (cb will be called with (err, res))

typings/algoliasearch.d.ts
```diff
@@ -0,0 +1,1252 @@
...
+     * @param queries index name, query and query parameters
+     * @param cb callback(err, res)
+     * https://github.com/algolia/algoliasearch-client-js#multiple-queries---multiplequeries
+     */
+    search(queries:{ indexName:string, query:string, options:IAlgoliaQueryParameters }, params:any, cb:Function);
+    /**
+     * Query on multiple index
+     * @param queries index name, query and query parameters
+     * @param params query parameter
+     * @param cb callback(err, res)
```
vvo: Can we add that every (asynchronous) method returns a Promise if not callback passed?

typings/algoliasearch.d.ts
```diff
@@ -0,0 +1,1252 @@
...
+     * https://github.com/algolia/algoliasearch-client-js#add-objects---addobjects
+     */
+    addObjects(objects:Object[], cb:Function);
+    /**
+     * Add or replace a specific object
+     * @param object
+     * @param cb(err, res)
+     * https://github.com/algolia/algoliasearch-client-js#update-objects---saveobjects
+     */
+    saveObject(object:Object, cb:Function);
```
vvo: cb is always optionnal, you get a Promise when no cb.

typings/algoliasearch.d.ts
```diff
@@ -0,0 +1,1252 @@
...
+     */
+    setSettings(settings:IAlgoliaIndexSettings, cb:Function);
+    /**
+     * Search in an index
+     * @param query string
+     * @param params query parameters
+     * return Promise
+     * https://github.com/algolia/algoliasearch-client-js#search-in-an-index---search
+     */
+    search(query:string, params:IAlgoliaQueryParameters):Promise<IAlgoliaResponse>;
```
vvo: I think for search we should only advise to use the form:

```js
search({params}, [cb]}
```

query, params is not deprecated but there's no real advantage of using it. Makes the API simpler

typings/algoliasearch.d.ts
```diff
@@ -0,0 +1,1252 @@
...
+    search(query:string, params:IAlgoliaQueryParameters, cb:Function);
+    /**
+     * Search in an index
+     * @param query query string
+     * @param params query parameter
+     * @param cb(err, res)
+     * @param err() error callback
+     * https://github.com/algolia/algoliasearch-client-js#search-in-an-index---search
+     */
+    search(query:string, params:IAlgoliaQueryParameters, cb:Function, err:Function);
```
vvo: no err function

CBaptiste:
will be done on definitelytyped instead

PR #362, opened 2016-11-28
fix(searchFacet): rename searchFacet to searchForFacetValues

vvo:


bobylito:
Documentation is not updated with the same way of describing the feature

vvo:
did it in a backward compat way, updated the jsDoc

bobylito:
👍 I think we're good to go. Since there are tests for both methods, it seems pretty safe to merge (and even release, no?)

vvo:
> I think we're good to go. Since there are tests for both methods, it seems pretty safe to merge (and even release, no?)

Let's do this yes

PR #365, opened 2016-12-12
fix(retry strategy): Shared appId host usage, retry on first host after 1 minute

vvo:
This commit adds another layer of HA by remembering the last known
work host for one minute.

The cache keys are either:
- domain + appId + browser => Browser localStorage
- process/pageLoad + appId => Node.js, Browser fallback

This is still a bit raw I think, have some review.

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -3,9 +3,15 @@ module.exports = AlgoliaSearchCore;
 var errors = require('./errors');
 var exitPromise = require('./exitPromise.js');
 var IndexCore = require('./IndexCore.js');
+var store = require('./store.js');
 
-// We will always put the API KEY in the JSON body in case of too long API KEY
+// We will always put the API KEY in the JSON body in case of too long API KEY,
+// to avoid query string being too long and failing in various conditions (our server limit, browser limit,
+// proxies limit)
 var MAX_API_KEY_LENGTH = 500;
+var RESET_TO_FIRST_HOST_TIMER =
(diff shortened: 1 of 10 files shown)


bobylito:
Is going back to the first host (the DSN in the case of Algolia default values) make sense when it comes to hosts provided by our users?

rayrutjes:
Shouldn't we retry after 5 minutes instead of 1 minute?

src/AlgoliaSearchCore.js
```diff
@@ -3,9 +3,11 @@ module.exports = AlgoliaSearchCore;
 var errors = require('./errors');
 var exitPromise = require('./exitPromise.js');
 var IndexCore = require('./IndexCore.js');
+var store = require('./store.js');
 
 // We will always put the API KEY in the JSON body in case of too long API KEY
 var MAX_API_KEY_LENGTH = 500;
+var RESET_TO_FIRST_HOST_TIMER = 60 * 60 * 1000;
```
bobylito: Isn't that an hour?

redox: Good catch 👍

vvo: Thx, updated to 5min

vvo:
> Shouldn't we retry after 5 minutes instead of 1 hour?

updated

vvo:
Added a unit test around retrying the first host

vvo:
- unit tests :white_check_mark: Chrome latest
- unit tests :white_check_mark: IE9
- unit tests :white_check_mark: IE10
- unit tests :white_check_mark: IE11

PR #366, opened 2016-12-13
fix(Node.js request strategy): differentiate connect from complete timeout

vvo:
Before this commit, Node.js implementation had only a complete timeout
of 30s. Without any connect timeout.

Now Node.js has a 2s connect timeout, 5s read timeout, 30s write
timeout as per specifications, cc @asurak

There's also an opportunity to add connect timeout handling to
frontend users. Should we? If so, what would be a good connect timeout
knowing that we already have lowered the "read" timeout for instance.


src/AlgoliaSearchCore.js
```diff
@@ -686,6 +710,13 @@ AlgoliaSearchCore.prototype._incrementCurrentHostIndex = function(hostType) {
   );
 };
 
+AlgoliaSearchCore.prototype._getTimeoutsForRequest = function(hostType, tries) {
+  return {
+    connect: this._timeouts.connect * tries,
```
bobylito: To my knowledge, timeouts gets double after each try. Here we have timeouts that are progressing linearly. Is it on purpose?

More details on what I understand if the connect timeout is 1:
 - here for each tries (3): 1, 2, 3
 - what I thought is "sorta" standard: 1, 2, 4

*update after talking with @clement-leprovost* It is the same in the swift counterpart :)

src/reactnative/builds/algoliasearch.reactnative.js
```diff
@@ -24,6 +24,12 @@ function algoliasearch(applicationID, apiKey, opts) {
 
   opts._ua = opts._ua || algoliasearch.ua;
 
+  opts.timeouts = opts.timeouts || {
+    connect: 2 * 1000,
+    read: 4 * 1000,
```
bobylito: Maybe this should have the same values as the swift client? https://github.com/algolia/algoliasearch-client-swift/blob/master/Source/AbstractClient.swift#L124

bobylito:
I don't have answers for your question though :D

src/reactnative/builds/algoliasearch.reactnative.js
ElPicador: It's 5 minutes on every other API clients. 
But maybe it'll make more sense to reduce it for the front-end implementations?

src/AlgoliaSearchCore.js
vvo: Thanks for investigating still :) This is the actual SPEC.

src/reactnative/builds/algoliasearch.reactnative.js
vvo: Yes, front-end facing API clients are meant to be used in situations with a low hitsPerPage. Not like a backend client, thus we can lower down the read timeout

vvo:
New version of JS client is going out today 10am CET

PR #446, opened 2017-01-04
Ensure a same user agent can not be added twice

rayrutjes:


src/AlgoliaSearchCore.js
```diff
@@ -154,7 +154,9 @@ AlgoliaSearchCore.prototype.setExtraHeader = function(name, value) {
 * @param algoliaAgent the agent to add
 */
 AlgoliaSearchCore.prototype.addAlgoliaAgent = function(algoliaAgent) {
-  this._ua += ';' + algoliaAgent;
+  if (this._ua.indexOf(';' + algoliaAgent) === -1) {
```
bobylito: Note for later, it would probably be best to store the agents in an array :)

bobylito:
Did you have the issue in one of your project? Would love to learn more about that?

src/AlgoliaSearchCore.js
rayrutjes: What would be the pros of that approach? Because that was my initial idea but I prefer the indexOf approach because we don't need to hold any reference, we just check inside the existing string.

bobylito: Oh and btw with this we can add a second time the JS client UA because we test against `;agent` not just `agent`.

bobylito: Humm it seems that we can add the JS client UA twice because it doesn't have the `;` in front.

bobylito: You're right. It's nicer but mostly useless; with an array you have to:
 - serialize the array at each request
 - finding if a UA is already set is marginally better in time (because we don't have much data)

rayrutjes: Yes that is true, I think this is an acceptable edge case for now given we don't use this function to initialize the UA.

PR #495, opened 2017-01-30
feat(x-algolia-agent): specify x-algolia-agent at search time

### LINKED ISSUE #493: Let the user specify a custom user-agent when a request is triggered
Currently the only way to add a new user agent is by using the method `addAlgoliaAgent` on the client instance. However autocomplete.js does not have access to the client but only to an index instance. To solve this, it would be perfect if we could set the user agent directly when the request is sent.

For now this is only required for searching (autocomplete.js)

## Proposal

Since it's only required for search and the ability not documented, we could implement a new method (to limit the complexity of the signature of the function):

```js
var maybeResults  = index.searchWithAlgoliaAgent('my new fancy agent', 'query');

index.searchWithAlgoliaAgent('my new fancy agent', 'query', (err, results) => {console.log(err, results);});
```

WDYT?

bobylito:
**Summary**

Currently there is no way to specify an algolia agent at search time. The parameter `x-algolia-agent` is just ignored.

**Result**

This PR let the user specify a new Algolia agent the parameter `x-algolia-agent` (like with the REST API) and it will be appended to the ones specified with `addAlgoliaAgent`.

Fixes #493

### DIFF
--- a/test/spec/common/client/addAlgoliaAgent.js
+++ b/test/spec/common/client/addAlgoliaAgent.js
@@ -2,7 +2,7 @@
 
 var test = require('tape');
 
-test('client.addAlgoliaAgent(algoliaAgent)', function(t) {
+test('AddAlgoliaAgent and custom search-time agent with x-algolia-agent', function(t) {
   t.plan(1);
 
   var fauxJax = require('faux-jax');
@@ -20,9 +20,11 @@ test('client.addAlgoliaAgent(algoliaAgent)', function(t) {
   // Ensure we de-duplicate by re-adding the same agent a second time.
   client.addAlgoliaAgent('And some other incredible agent');
 
-  index.search('algolia agent');
+  index.search('algolia agent', {
+    'x-algolia-agent': 'the other agent'
+  });
 
-  var expectedAgent = fixture.algoliasearch.ua + ';And some other incredible agent';
+  var expectedAgent = fixture.algoliasearch.ua + ';And some other incredible agent;the other agent';
 
   fauxJax.once('request', function(req) {
     var agent = process.browser ?
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -168,6 +168,7 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
   var requestDebug = require('debug')('algoliasearch:' + initialOpts.url);
 
   var body;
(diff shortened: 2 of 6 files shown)


test/spec/common/client/addAlgoliaAgent.js
```diff
@@ -40,3 +40,44 @@ test('client.addAlgoliaAgent(algoliaAgent)', function(t) {
...
+
+  var createFixture = require('../../../utils/create-fixture');
+  var fixture = createFixture();
+
+  var client = fixture.client;
+  var index = fixture.index;
+
+  client.addAlgoliaAgent('And some other incredible agent');
+
+  // Ensure we de-duplicate by re-adding the same agent a second time.
```
vvo: Since this is already tested in another test, can we remove it?

test/spec/common/client/addAlgoliaAgent.js
```diff
@@ -40,3 +40,44 @@ test('client.addAlgoliaAgent(algoliaAgent)', function(t) {
...
+  client.addAlgoliaAgent('And some other incredible agent');
+
+  index.search('algolia agent', {
+    'x-algolia-agent': 'the other agent'
+  });
+
+  var expectedAgent = fixture.algoliasearch.ua + ';And some other incredible agent';
```
vvo: I would have expected the agent to be:

fixture.algoliasearch.ua + ';the other agent'. I cannot see where do we test for the presence of "the other agent"

bobylito: #fail :/

bobylito:
Need more changes :)

bobylito:
Ok so I gave on trying the generic approach. I added a specific parameter for adding a new additional UA in the internal API and only provided an user parameter for search (which is the use case we needed). @vvo

bobylito:
I should something for search for facet values maybe 🤔

vvo:
Here's what I get when doing `npm test` in this branch:

image

I guess travis is not showing it right maybe because of weird CI issue on colors..

vvo:
(most probably since I started using tap-spec thing)

bobylito:
Of course, node specific implementation of `_computeRequestHeaders` :)

test/spec/common/client/addAlgoliaAgent.js
bobylito: Sure :) I remember the discussion we had on this kind of scenarios for tests :)

PR #496, opened 2017-02-01
fix(algolia agent): Fix parameter name for algolia agent

bobylito:
**Summary**

The algolia agent can be set with x-algolia-agent in the parameters used with the search method. However additionalUA is more explicit.

**Result**

```js
index.search('search it', {addtionalUA: 'super search agent'});
```

This adds a `super search agent` to the `x-algolia-agent` header.


vvo:
Cleared the cache on travis and relaunched

PR #531, opened 2017-03-08
[WIP] Provide es module build

rayrutjes:
Same as https://github.com/algolia/algoliasearch-helper-js/pull/482


rayrutjes:
I have yet one warning that needs to be addressed for the server build:
```
'default' is not exported by 'node_modules/rollup-plugin-node-builtins/src/es6/http.js'
https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
commonjs-proxy:node_modules/rollup-plugin-node-builtins/src/es6/http.js (1:173)
1: import * as http from "node_modules/rollup-plugin-node-builtins/src/es6/http.js"; export default ( http && http['default'] ) || http;
```

rayrutjes:
The server build needs some testing to be sure.

rayrutjes:
Please note that the server ES6 build has not been tested yet. Please do not rush this ;)

rayrutjes:
Given that in the client we have different builds for browser and the client, it would currently mean creating 2 separate es2015 module. Problem being there should only be one.

Given we can't leverage tree shaking and that we can not have 1 single es2015 module that covers all the use cases, I'll close this.

For ES6 projects, users will have to choose wether to user the `browser` field or not depending on the target platform and transform es5 exports to es6 exports.

vvo:
Even if it looks like @rayrutjes has been struggling all alone with this, we have been following this internally together. We do want to make the JS client modern by providing more ES6/modules goodness. The JS community around tree shaking/modules/es6/export/import is still very young but most tooling is getting ready for the big jump. We will do it more probably this year.

oncletom:
Hello,

I was wondering if there was an issue to follow the work on ESModules? I'm especially interested in importing modules without a bundler (like webpack and browserify).

Haroenv:
@oncletom, there's currently no issue opened about this, but there's general modernisation work we will do which should include es modules at some point, it's just that the experiment we did here proved that we had some extra complexity we didn't expect (especially with node for example). 

However, you can use it without bundling by using the UMD bundle: 

```html
<script src="https://cdn.jsdelivr.net/npm/algoliasearch@3.30.0/dist/algoliasearch.min.js"></script>
```

PR #533, opened 2017-03-10
Check if 'process' exists prior to using it in browser build

rayrutjes:
This pull request will make sure the `process` variable is set prior to using it.

This is currently not an issue regarding the existing browser builds because `envify` will take care if initializing the `process` variable.

That being said, if some user wants to bundle the client in UMD format for example, then he will need to specify the following in his build:

```js
// Rollup example
replace({
      'process.env.NODE_ENV': JSON.stringify('production'),
      'process.env.RESET_APP_DATA_TIMER': 120000,
    }),
```

Unless this is indeed the expected behaviour, this PR solves the issue.

I'm wondering about the `NODE_ENV` though. Maybe we do want the users to override it that way, it is kinda common. With this PR, we can not simply "replace" it because we first check the `process`.

Happy to have your feedback on this.

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -10,7 +10,7 @@ var store = require('./store.js');
 // proxies limit)
 var MAX_API_KEY_LENGTH = 500;
 var RESET_APP_DATA_TIMER =
-  process.env.RESET_APP_DATA_TIMER && parseInt(process.env.RESET_APP_DATA_TIMER, 10) ||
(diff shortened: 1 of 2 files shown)


rayrutjes:
So it appears that keeping the code like it is is a good practise.

Regarding my previous example, as a workaround, the solution is to replace `process.env` with an object containing only `NODE_ENV`.

```js
// In webpack config
// ...
plugins: [
    new webpack.DefinePlugin({
      'process.env': {
        'NODE_ENV': '"production"'
      }
    })
  ]
//...
```

```js
// In rollup config
// ...
replace({
      'process.env': JSON.stringify({
        'NODE_ENV': 'production'
      })
    }),
//...
```

As long as the `process.env` is defined, the code will work nicely.

PR #534, opened 2017-03-10
feat(api keys): rename all *userKey to *apiKey

vvo:
with backward compat and depreciation messages


redox:
I'm always confused when it comes to acronyms... whether we should camelCase them or not :/ I think I would have written `listAPIKeys()` instead of `listApiKeys` but no strong opinion.

vvo:
> I'm always confused when it comes to acronyms... whether we should camelCase them or not :/ I think I would have written listAPIKeys() instead of listApiKeys() but no strong opinion.

I would also prefer to always write api as API even in this case, so listAPIKeys(). Wonder if it's a real struggle or just another OCD.

Our chosen spec was to use listApiKeys form from what I read and see what's implemented already in other clients: https://github.com/algolia/algoliasearch-client-php/commit/c9bdad396d871fbc11022c1fa82e43d6b0333bcb

Still using API in comments.

rayrutjes:
Same feeling @redox I think it all comes down to personal prefs and consistency in projects.
I use both notations depends on the project. Personally I prefer the Api approach, because when you need to convert to first char to lower, you end up with aPI otherwise.

vvo:
Very common acronyms (i.e. not OUR acronyms, like applicationID, we ditched this in favour of applicationId or appId) should stay capitalized from my POV. React thinks the same: https://facebook.github.io/react/docs/dom-elements.html#dangerouslysetinnerhtml

image

But at the same time, they have:

image

So is the rule:
- when acronym is in leading position: lowercase?
- when acronym is in trailing position: uppercase?

Haroenv:
htmlFor comes from the DOM API https://developer.mozilla.org/en-US/docs/Web/API/HTMLLabelElement/htmlFor

vvo:
this ships on monday so we can follow the deployment

PR #546, opened 2017-04-23
Use Array.isArray directly

krienow:
**Summary**

isarray is deprecated. We can use the built-in function Array.isArray directly.

**Result**

Removed isarray dependency and replaced all usages with Array.isArray.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
     "foreach": "^2.0.5",
     "global": "^4.3.0",
     "inherits": "^2.0.1",
-    "isarray": "^2.0.1",
     "load-script": "^1.0.0",
     "object-keys": "^1.0.11",
     "querystring-es3": "^0.2.1",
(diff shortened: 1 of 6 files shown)


Haroenv:
We still support ie8 unfortunately, so this can't be used directly until we drop that. Cc @vvo

krienow:
What about a polyfill instead of another dependency?

```
if (!Array.isArray) {
  Array.isArray = function(arg) {
    return Object.prototype.toString.call(arg) === '[object Array]';
  };
}
```

Source: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

Haroenv:
`isarray` is three lines: https://unpkg.com/isarray@2.0.1

```js
var toString = {}.toString;

module.exports = Array.isArray || function (arr) {
  return toString.call(arr) == '[object Array]';
};
```

vvo:
Hi @krienow, is your issue about the fact that installing algoliasearch displays a warning about isarray being deprecated?

If so I am sorry that's the case, this module (algoliasearch) will most probably be updated this year to get a facelift including using less dependencies like this one.

If that's the only struggle you have right now, please close this issue but keep sending feedback. If you have a more serious issue because of using isarray, let us know.

krienow:
Yes, it was about the warning - nothing important. Keep up the good work! 😄

PR #552, opened 2017-05-18
fix(partialUpdateObjects): allow createIfNotExists

### LINKED ISSUE #551: Offer partialUpdateObjectNoCreate action in partialUpdateObjects function
Feature for this [function](https://github.com/algolia/algoliasearch-client-javascript/blob/b7bd36b357ca20a23e90ce21375dacb7bd878210/src/Index.js#L140). Asked by a customer.

Right now, the action in the request can only be `partialUpdateObject`. Would be good to add a way to opt in to use `partialUpdateObjectNoCreate`.
In Swift, we simply added a param `createIfNotExists` to the function `partialUpdateObjects`, and then did:
```
let action = createIfNotExists ? "partialUpdateObject" : "partialUpdateObjectNoCreate"
```

This should work `partialUpdateObjects`, not only for `partialUpdateObject` as is the case right now.

vvo:
Much like partialUpdateObject, allow createIfNotExists

fixes #551

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -123,7 +123,12 @@ Index.prototype.partialUpdateObject = function(partialObject, createIfNotExists,
 *  error: null or Error('message')
 *  content: the server answer that updateAt and taskID
 */
-Index.prototype.partialUpdateObjects = function(objects, callback) {
+Index.prototype.partialUpdateObjects = function(objects, createIfNotExists, callback) {
+  if (arguments.length === 1 || typeof createIfNotExists === 'function') {
+    callback = createIfNotExists;
+    createIfNotExists = true;
+  }
+
   var isArray = require('isarray');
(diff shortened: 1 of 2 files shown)


vvo:
Oh Wow travis passed just like that. good

src/Index.js
```diff
@@ -123,7 +123,12 @@ Index.prototype.partialUpdateObject = function(partialObject, createIfNotExists,
 *  error: null or Error('message')
 *  content: the server answer that updateAt and taskID
 */
-Index.prototype.partialUpdateObjects = function(objects, callback) {
+Index.prototype.partialUpdateObjects = function(objects, createIfNotExists, callback) {
+  if (arguments.length === 1 || typeof createIfNotExists === 'function') {
```
Haroenv: good workaround that keeps the old behaviour!

PR #564, opened 2017-06-26
refact(rules): Adapt to latest JSON schema for Query Rules

clement-leprovost:
**Summary**

`if`/`then` have been renamed to `condition`/`consequence` to avoid conflicts with reserved keywords in strongly typed clients that want to enforce the Rules schema.

**Result**

This PR adapts the integration tests accordingly.


vvo:
$if
$then

would have worked too :D

Issue #569, opened 2017-07-16
deprecate methods in V3 that will be removed in v4


Haroenv:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

All methods exist

**What is the expected behavior?**

the following methods should work exactly the same as before, but should have a deprecation message and an entry in the wiki: 

- [ ] addObject
- [ ] saveObject

There are more to come, but singular/plural + addObject / saveObject confusion should be handled in the least

PR #570, opened 2017-07-16
feat(deprecation): use console.warn to be more visible

Haroenv:
**Summary**

Use console.warn rather than console.log This is the same thing as Facebook uses for React, and is more visible than a simple console.log. It is [supported](https://developer.mozilla.org/en-US/docs/Web/API/Console/warn#Browser_compatibility) in IE8, so that should be fine

Slightly related to #569

**Result**

Deprecation warnings are now in yellow and marked as a warning so we can remove them with more ease later.


vvo:
👌  I see no issue with the move. Thanks a lot.

PR #572, opened 2017-07-18
feat(headers): add functions to remove and get extra headers

### LINKED ISSUE #571: No way to add and remove headers
In some dashboard situations, @alexandremeunier would like to add and remove headers easily.

Right now we only have setExtraHeaders and it manipulates an array. So you can't do:

```js
client.setExtraHeader('...', '...');
client.unsetExtraHeader('...');
// or client.setExtraHeader('name', null); // this just adds currently
```

1. It should manipulate a map
2. It should have an unset

Can we make a patch in V3 for this? Thanks (that would be your first client release @Haroenv :D)

In V4 we don't have to think about this first, we will add it in some other way. (to be discussed)

Haroenv:
**Summary**

There's no way to check an extra header or remove one, so that needed to be done

**Result**

```js
client.setExtraHeader('X-cool-header','hello there');
client.getExtraHeader('X-cool-header'); //hello there
client.deleteExtraHeader('X-cool-header');
client.getExtraHeader('X-cool-header'); //undefined
```

Do I need to add more documentation for this somewhere?

cc @alexandremeunier 

closes #571

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -111,7 +111,7 @@ function AlgoliaSearchCore(applicationID, apiKey, opts) {
   this.hosts.read = map(this.hosts.read, prepareHost(protocol));
   this.hosts.write = map(this.hosts.write, prepareHost(protocol));
 
-  this.extraHeaders = [];
+  this.extraHeaders = {};
 
   // In some situations you might want to warm the cache
   this.cache = opts._cache || {};
@@ -142,9 +142,25 @@ AlgoliaSearchCore.prototype.initIndex = function(indexName) {
 * @param value the header field value
 */
 AlgoliaSearchCore.prototype.setExtraHeader = function(name, value) {
-  this.extraHeaders.push({
-    name: name.toLowerCase(), value: value
-  });
+  this.extraHeaders[name.toLowerCase()] = value;
+};
+
+/**
+* Get the value of an extra HTTP header
+*
+* @param name the header field name
+*/
+AlgoliaSearchCore.prototype.getExtraHeader = function(name) {
+  return this.extraHeaders[name.toLowerCase()];
+};
+
+/**
+* Remove an extra field from the HTTP request
+*
+* @param name the header field name
+*/
+AlgoliaSearchCore.prototype.unsetExtraHeader = function(name) {
+  delete this.extraHeaders[name.toLowerCase()];
 };
 
 /**
@@ -486,11 +502,9 @@ AlgoliaSearchCore.prototype._computeRequestHeaders = function(additionalUA, with
     requestHeaders['x-algolia-tagfilters'] = this.securityTags;
   }
 
-  if (this.extraHeaders) {
-    forEach(this.extraHeaders, function addToRequestHeaders(header) {
-      requestHeaders[header.name] = header.value;
-    });
-  }
+  forEach(this.extraHeaders, function addToRequestHeaders(value, key) {
+    requestHeaders[key] = value;
+  });
 
   return requestHeaders;
 };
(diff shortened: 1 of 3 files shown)


test/spec/common/client/interface.js
```diff
@@ -20,25 +20,25 @@ test('AlgoliaSearch client API spec', function(t) {
 
   var expectedProperties = [
     'addAlgoliaAgent',
-    'addQueryInBatch',
     'addApiKey',
+    'addQueryInBatch',
     'addUserKey',
     'addUserKeyWithValidity',
     'batch',
-    'updateApiKey',
```
Haroenv: I sorted them alphabetically, Ince that seemed to be what they used to be sorted in once

test/spec/common/client/setExtraHeader.js
```diff
@@ -3,7 +3,7 @@
 var test = require('tape');
 
 test('client.setExtraHeader(key, value)', function(t) {
-  t.plan(4);
+  t.plan(11);
```
Haroenv: I don't know why this was originally 4 since there were already 6 tests

Haroenv:
As noted in https://tools.ietf.org/html/rfc7230#section-3.2.2, http://greenbytes.de/tech/webdav/rfc2616.html#message.headers and https://stackoverflow.com/questions/39735299/is-an-http-client-allowed-to-send-multiple-headers-with-the-same-name you can't really have multiple headers with the same name, if that's the case, you should send an array, instead of calling setExtraHeader multiple times

test/spec/common/client/interface.js
bobylito: Good call :)

src/AlgoliaSearchCore.js
```diff
@@ -142,9 +142,25 @@ AlgoliaSearchCore.prototype.initIndex = function(indexName) {
...
+AlgoliaSearchCore.prototype.getExtraHeader = function(name) {
+  return this.extraHeaders[name.toLowerCase()];
+};
+
+/**
+* Remove an extra field from the HTTP request
+*
+* @param name the header field name
+*/
+AlgoliaSearchCore.prototype.deleteExtraHeader = function(name) {
```
vvo: Wondering about delete vs unset (since we use set), any arguments?

src/AlgoliaSearchCore.js
```diff
@@ -458,6 +474,18 @@ AlgoliaSearchCore.prototype._getSearchParams = function(args, params) {
   return params;
 };
 
+// jQuery's version of isEmpty
```
vvo: Can you put a link to the source? with proper sha and maybe copyright

src/AlgoliaSearchCore.js
```diff
@@ -486,9 +514,9 @@ AlgoliaSearchCore.prototype._computeRequestHeaders = function(additionalUA, with
     requestHeaders['x-algolia-tagfilters'] = this.securityTags;
   }
 
-  if (this.extraHeaders) {
-    forEach(this.extraHeaders, function addToRequestHeaders(header) {
-      requestHeaders[header.name] = header.value;
+  if (!isEmpty(this.extraHeaders)) {
```
vvo: Would have used object keys but it's not available in IE8 :)

vvo:
> Do I need to add more documentation for this somewhere?

Maybe, check doc repository and search for setExtraHeader

src/AlgoliaSearchCore.js
Haroenv: Unfortunately, I thought the same 😭

Haroenv:
`setExtraHeader` is not documented, so there’s no need to add documentation for these methods either 😄

src/AlgoliaSearchCore.js
vvo: In the end, we do not even need to check isEmpty or not, since the forEach will not iterate on it if it has no properties, let's remove this code?

Haroenv: Great call ❤️

Haroenv:
Merging, since the build fails at the same IE8 point as before

PR #573, opened 2017-07-20
chore(deprecation): add deprecation message to all methods that will be removed in v4

Haroenv:
**Summary**

These methods will be served by their plural form in v4 and should thus be deprecated.

**Result**

All methods that will be removed are now deprecated, with a wiki page about them 

todo: 

- [x] write wiki pages

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -28,7 +28,7 @@ inherits(Index, IndexCore);
 *  error: null or Error('message')
 *  content: the server answer that contains 3 elements: createAt, taskId and objectID
 */
-Index.prototype.addObject = function(content, objectID, callback) {
+Index.prototype.addObject = deprecate(function(content, objectID, callback) {
   var indexObj = this;
 
   if (arguments.length === 1 || typeof objectID === 'function') {
@@ -46,7 +46,7 @@ Index.prototype.addObject = function(content, objectID, callback) {
     hostType: 'write',
     callback: callback
   });
-};
+}, deprecatedMessage('index.addObject(obj)', 'index.addObjects([obj])'));
 
 /*
 * Add several objects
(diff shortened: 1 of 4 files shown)


src/Index.js
```diff
@@ -46,7 +46,7 @@ Index.prototype.addObject = function(content, objectID, callback) {
     hostType: 'write',
     callback: callback
   });
-};
+}, deprecatedMessage('index.addObject()', 'index.addObjects()'));
```
vvo: Should we add addObjects([object]) as the replacement? (Same everywhere)

src/Index.js
```diff
@@ -493,7 +493,10 @@ Index.prototype.ttAdapter = function(params) {
       cb(content.hits);
     });
   };
-};
+},
+'ttAdapter is deprecated, there is no replacement planned, but open an issue on\n' +
```
vvo: Message should say: ttAdapter is no more necessary, have a look at autocomplete.js (link)

src/Index.js
Haroenv: Ah, didn't think of that, sounds like a good idea, but in the new version, would we ask for people to do: 

```js
index.addObjects([object]);

// or

index.addObjects(object);
```

I assumed you meant the second method, but I think I prefer the first code and consistency-wise

vvo: It's a tough one, I would say we should accept only the array for consistency and precision but still have good errors in term of usage.

variadic, boolean traps, magic obj => [obj] decisions are most of the time bad choices we regret later on.

Haroenv: perfect, that means that behaviour won't change between v3 and v4 of the plural forms. 

I'm *slightly* wondering if there might be a performance difference between the single form and the plural form with a single item, WDYT (maybe someone from core team?)

vvo: This should be completely equivalent

Haroenv:
It should be good to go now, I wrote the wiki pages too for those added here, but I see that there are some older ones that don't have an anchor yet there. Will add those later.

wiki: https://github.com/algolia/algoliasearch-client-javascript/wiki/Deprecated

I also fixed `deprecatedMessage` a bit so that it works with parameters in the function call

PR #575, opened 2017-07-24
chore(deprecation): remove deprecation message from singular methods

Haroenv:
**Summary**

The plural and singular  methods are still useful, and will be kept in v4

**Result**

methods no longer deprecated

* getObject
* addObject
* saveObject
* partialUpdateObject
* deleteObject


vvo:
To anyone digging the commit history: good luck 🤣 

Seriously though: this is good, thanks

PR #577, opened 2017-07-27
Places API V4 proposal

vvo:
Little changes to the examples and actual naming around getObject getObjects => getPlace getPlaces.

### DIFF
--- a/API.md
+++ b/API.md
@@ -29,12 +29,12 @@ index.requester // clearIndex, headers, timeouts
 ```js
 import { initPlaces } from 'algoliasearch';
 
-const index = initPlaces(appId, apiKey);
+const places = initPlaces(appId, apiKey);
 
-index.search(params);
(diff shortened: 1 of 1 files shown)


vvo:
I wonder if this makes sense to use getPlace and getPlaces rather than getObject. Maybe that's not so good idea because then people will not easily do the link between places and indexes when they want to mix both.

bobylito:
> I wonder if this makes sense to use getPlace and getPlaces rather than getObject. Maybe that's not so good idea because then people will not easily do the link between places and indexes when they want to mix both.

On the other hand it clearly shows that those are different. Also getPlaces is more semantically accurate. That's a good name IMHO

vvo:
Thx @bobylito for feedback on this :)

Haroenv:
That's a great idea 😄

PR #603, opened 2017-09-26
feat(deleteBy): add new deleteBy method

Haroenv:
**Summary**

This replaces `deleteByQuery`, as it's now an API feature

see (internal) algolia/doc#782

**Result**

new method `deleteBy` on Index, `deleteByQuery` deprecated (and ☁️ gone in v4)

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -383,7 +383,7 @@ Index.prototype.deleteBy = function(params, callback) {
     method: 'POST',
     url: '/1/indexes/' + encodeURIComponent(indexObj.indexName) + '/deleteByQuery',
     body: {params: indexObj.as._getSearchParams(params, '')},
-    hostType: 'read',
+    hostType: 'write',
     callback: callback
   });
 }
--- a/src/Index.js
+++ b/src/Index.js
@@ -366,6 +366,13 @@ Index.prototype.deleteByQuery = deprecate(function(query, params, callback) {
 /**
 * Delete all objects matching a query
 *
+* the query parameters that can be used are: 
+* - query
+* - filters (numeric, facet, tag)
+* - geo
+*
+* you can not send an empty query or filters
+*
 * @param params the optional query parameters
 * @param callback (optional) the result callback called with one argument
 *  error: null or Error('message')
@@ -375,7 +382,7 @@ Index.prototype.deleteBy = function(params, callback) {
   return this.as._jsonRequest({
     method: 'POST',
     url: '/1/indexes/' + encodeURIComponent(indexObj.indexName) + '/deleteByQuery',
-    body: {params: params},
+    body: {params: indexObj.as._getSearchParams(params, '')},
     hostType: 'read',
     callback: callback
   });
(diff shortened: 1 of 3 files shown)


src/Index.js
```diff
@@ -360,7 +361,25 @@ Index.prototype.deleteByQuery = function(query, params, callback) {
...
+/**
+* Delete all objects matching a query
+*
+* @param params the optional query parameters
+* @param callback (optional) the result callback called with one argument
+*  error: null or Error('message')
+*/
+Index.prototype.deleteBy = function(params, callback) {
+  var indexObj = this;
+  return this.as._jsonRequest({
```
Haroenv: note that this is __not__ a variadic call, to align with v4 already.

src/Index.js
```diff
@@ -360,7 +361,25 @@ Index.prototype.deleteByQuery = function(query, params, callback) {
       callback(err);
     }, client._setTimeout || setTimeout);
   }
-};
+}, deprecatedMessage('index.deleteByQuery()', 'index.deleteBy()'));
```
vvo: You will have to add an entry in the wiki and explain the ({query})

src/Index.js
```diff
@@ -360,7 +361,25 @@ Index.prototype.deleteByQuery = function(query, params, callback) {
...
+*
+* @param params the optional query parameters
+* @param callback (optional) the result callback called with one argument
+*  error: null or Error('message')
+*/
+Index.prototype.deleteBy = function(params, callback) {
+  var indexObj = this;
+  return this.as._jsonRequest({
+    method: 'POST',
+    url: '/1/indexes/' + encodeURIComponent(indexObj.indexName) + '/deleteByQuery',
```
vvo: Can you add a unit test (takes 5 min, go into spec/anySimpleMethod)

src/Index.js
```diff
@@ -360,7 +361,25 @@ Index.prototype.deleteByQuery = function(query, params, callback) {
...
+    body: {params: params},
```
Haroenv: this should be stringified, right now it's an object.

src/Index.js
Haroenv: I added a unit test, but it doesn't fail when it should 🤔

test/spec/common/index/test-cases/deleteBy.js
```diff
@@ -0,0 +1,96 @@
...
+      body: {
+        params: 'query=some%22q',
+      },
+    },
+  },
+  {
+    object: 'index',
+    methodName: 'deleteBy',
+    testName: 'index.deleteBy({query, filters}, cb)',
+    // this test should fail but it doesn't, so I don't trust the new tests 😭
```
Haroenv: it doesn't enforce the error so I'm not sure what happens

vvo: Why should it fails?

Haroenv: I added `ridiculous:'mistake'` to the arguments to test if it was working, but it's not in the params and the test doesn't fail

PR #608, opened 2017-10-11
chore(deprecation): deprecate waitTask --> waitForCompletion

Haroenv:
We renamed this one in v4 because it's not really proper English

- [x] [wiki entry](https://github.com/algolia/algoliasearch-client-javascript/wiki/Deprecated#indexwaittask)
- [x] alias in v3
- [x] v4 implementation

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -532,7 +532,7 @@ Index.prototype.ttAdapter = deprecate(function(params) {
 *  error: null or Error('message')
 *  content: the server answer that contains the list of results
 */
-Index.prototype.waitTask = function(taskID, callback) {
+Index.prototype.waitTask = deprecate(function(taskID, callback) {
   // wait minimum 100ms before retrying
   var baseDelay = 100;
   // wait maximum 5s before retrying
@@ -583,7 +583,7 @@ Index.prototype.waitTask = function(taskID, callback) {
       callback(err);
     }, client._setTimeout || setTimeout);
   }
-};
+}, deprecatedMessage('index.waitTask()', 'index.waitForCompletion()'));
 
 /*
 * This function deletes the index content. Settings and index specific API keys are kept untouched.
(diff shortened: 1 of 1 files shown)


vvo:
Where's waitForCompletion implemented?

Haroenv:
It's not implemented yet in v3, forgot to push that

vvo:
May need to update tests too.

Is waitForCompletion the new waitTask for all API clients?

Haroenv:
I think it was originally @codeharmonics suggesting this change

Haroenv:
Java has `waitForCompletion`, Scala `waitFor`, other ones have `waitTask`, `wait_task`

rayrutjes:
I feel like we should try to keep some consistency.
If we are to rename methods like these, we should validate them with the API maintainers and implement the change everywhere?

rayrutjes:
Java and Scala are the exceptions here if I'm not mistaken.

redox:
I agree with @rayrutjes, not sure we want to introduce a new wording here. This method should anyway never be used for something else that testing purposes.

codeharmonics:
@clement-leprovost started a discussion about renaming waitTask. Everybody agreed that it needed to be more explicit - wait for the task to do what???  Some suggested waitForTask, waitForTaskToComplete, or waitForCompletion.

rayrutjes:
@codeharmonics whatever is best indeed. Some of the suggestions really are more explicit than today.
However, I think that renaming single methods like this will create a lot of noise.
Maybe we should start creating a document listing all `methods` / `concepts` etc so that we can eventually all go in the same direction when releasing major versions.

@Haroenv also has some nice ideas around `Requester` etc which are probably worth discussing / implementing in other clients in the future.

codeharmonics:
Doc team will be revisiting our API client methods, adding exact parameters and return results, as well as more code samples. We can start creating a list of name changes. But I'm curious how we manage interface changes? Can old clients still use the existing names? Are new clients encouraged or even required to use the new names?

Haroenv:
If this is too contentious of a name change I can keep it like it is, but I only made this PR because we discussed it on Slack and it was on my TODO

Haroenv:
Reverting this. We will reconsider this kind of naming on a global level at a later stage

PR #617, opened 2017-11-10
feat(synonyms): exportSynonyms and exportRules

Haroenv:
in v4 we will rework this under algoliasearch-extras with as much control as needed, similar for browseAll

**Summary**

exportSynonyms and exportRules will get all of the synonyms / rules on an index

**Result**

both implemented naively with the only way to control the flow being specifying hitsPerPage

**Test plan**

```js
index.batchRules([someRules]);
index.searchRules(hitsPerPage?: number).then(saveToFile);
// same for synonyms
```

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -636,7 +636,7 @@ Index.prototype.searchSynonyms = function(params, callback) {
   });
 };
 
-function exportData(methodName, _hitsPerPage) {
+function exportData(methodName, _hitsPerPage, callback) {
   function search(page, _previous) {
     var options = {
       page: page || 0,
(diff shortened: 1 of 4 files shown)


Haroenv:
ugh, we don't have enough synonyms on the test app? https://travis-ci.org/algolia/algoliasearch-client-javascript/builds/301357682#L566

src/Index.js
```diff
@@ -745,6 +773,10 @@ Index.prototype.searchRules = function(params, callback) {
   });
 };
 
+Index.prototype.exportRules = function(hitsPerPage) {
```
vvo: Need to pass callback down the line too no? Otherwise here it does not supports callback form.

Haroenv: I was thinking to have this as a "new functions don't have callbacks anymore", but probably better to keep it indeed. Problem is that we use promises internally for this, so you'll still need support (maybe okay with the polyfill we have)

src/Index.js
```diff
@@ -636,6 +636,45 @@ Index.prototype.searchSynonyms = function(params, callback) {
...
+  });
+}
+
+/**
+ * Retrieve all the synonyms in an index
+ * @param [number=100] hitsPerPage The amount of synonyms to retrieve per batch
+ * @param [function] callback will be called after all synonyms are retrieved
+ */
+Index.prototype.exportSynonyms = function(hitsPerPage, callback) {
+  return exportData(this.searchSynonyms, hitsPerPage, callback);
```
Haroenv: It's not exporting 0, 1, 2, and everything is in it twice, and I'm completely lost, @vvo do you mind taking a look?

Haroenv: solved

PR #621, opened 2017-11-20
chore(sizing): add bundlesize

Haroenv:
**Summary**

This will test on PR what the built size of the library is.
This is also used in instantsearch.js

**Result**

Now I just added 1kb more than the current size so we don't accidently go above it by a lot and we explicitly know which file has which limit.


Haroenv:
I'm wondering if we should put the lint and this test before the functional one: 

```patch
diff --git a/scripts/test-ci b/scripts/test-ci
index aaed248..b35ab06 100755
--- a/scripts/test-ci
+++ b/scripts/test-ci
@@ -11,6 +11,12 @@ echo "CI test"
 echo "CI test: node"
 ./scripts/test-node

+echo "CI test: lint"
+./scripts/lint
+
+echo "Testing bundle size"
+bundlesize
+
 echo "CI test: browser"
 ./scripts/test-browser

@@ -20,9 +26,3 @@ else
   echo "CI test: parse-cloud"
   ./scripts/test-parse-cloud
 fi
-
-echo "CI test: lint"
-./scripts/lint
-
-echo "Testing bundle size"
-bundlesize
```

vvo:
> I'm wondering if we should put the lint and this test before the functional one:

I would say having functional tests are more important than knowing the linting and size fails (i.e. I still want to be able to have the functional tests running even if I added a widget)

How does it works when the bundle size goes 1kb over ? Can we "accept it" somehow? Or do we have to raise the limit?

Haroenv:
Accepting would be raising the limit in package.json @vvo, since it's a "staying" thing

PR #623, opened 2017-11-20
chore(test): remove IE8 on zuul

Haroenv:
**Summary**

IE8 always failing, which causes us to always ignore, which makes this test useless.

**Result**

Removed Zuul testing for IE8

I also removed the `browzers` dependency because it makes more sense to manage this testing inline

### DIFF
--- a/zuul.config.js
+++ b/zuul.config.js
@@ -33,7 +33,56 @@ if (process.env.CI === 'true') {
   };
 }
 
-var browsers = require('browzers');
+var browsers = {
+  all: [{
+    name: 'chrome',
+    version: '-1..latest',
+    platform: 'Windows 10'
+  }, {
+    name: 'firefox',
+    version: '-1..latest',
+    platform: 'Windows 10'
+  }, {
+    name: 'internet explorer',
+    version: '9..11'
+  }, {
+    name: 'safari',
+    version: '-3..latest'
+  }, {
+    name: 'iphone',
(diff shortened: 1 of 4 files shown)


Haroenv:
Ugh, I think IE11 is also a bit broken 😢 https://travis-ci.org/algolia/algoliasearch-client-javascript/builds/304872047#L1448

Haroenv:
ugh

zuul.config.js
```diff
@@ -33,7 +33,56 @@ if (process.env.CI === 'true') {
...
+    name: 'microsoftedge',
+    version: 'latest'
+  }],
+  pullRequest: [{
+    name: 'chrome',
+    version: 'latest',
+    platform: 'Windows 10'
+  }, {
+    name: 'internet explorer',
+    version: ['9']
```
Haroenv: We will need to test manually for other versions, but I think it's ok

zuul.config.js
```diff
@@ -33,7 +33,56 @@ if (process.env.CI === 'true') {
...
+    name: 'chrome',
+    version: '-1..latest',
+    platform: 'Windows 10'
+  }, {
+    name: 'firefox',
+    version: '-1..latest',
+    platform: 'Windows 10'
+  }, {
+    name: 'internet explorer',
+    version: '9..10'
```
Haroenv: @vvo should this be done on 9 alone too then 🤔

PR #643, opened 2017-12-13
feat(mcm): add methods for multi cluster management

Haroenv:
**Summary**

Adding the MCM methods

**Result**

Also have added the option to add `headers` to `_jsonRequest` and changed it into an object-accepting function for "future-proofing" it


src/AlgoliaSearch.js
```diff
@@ -487,6 +487,159 @@ AlgoliaSearch.prototype.batch = function(operations, callback) {
...
+/**
+ * Get the top userIDs
+ *
+ * (the callback is the second argument)
+ *
+ * @return {Promise|undefined} Returns a promise if no callback given
+ * @example
+ * client.getTopUserID();
+ */
+AlgoliaSearch.prototype.getTopUserID = function(_, callback) {
```
vvo: why not just func(callback)? why force to do func(null, cb)?

src/AlgoliaSearch.js
```diff
@@ -487,6 +487,159 @@ AlgoliaSearch.prototype.batch = function(operations, callback) {
...
+  return this._jsonRequest({
+    method: 'GET',
+    url: '/1/clusters/mapping/top',
+    hostType: 'write',
```
vvo: read?

src/AlgoliaSearch.js
```diff
@@ -487,6 +487,159 @@ AlgoliaSearch.prototype.batch = function(operations, callback) {
...
+ * client.getUserID({ userID: 'some-user' });
+ */
+AlgoliaSearch.prototype.getUserID = function(data, callback) {
+  if (!data.userID) {
+    throw new errors.AlgoliaSearchError('You have to provide a userID', {debugData: data});
+  }
+  return this._jsonRequest({
+    method: 'GET',
+    url: '/1/clusters/mapping/' + data.userID,
+    hostType: 'write',
```
vvo: read?

src/AlgoliaSearch.js
```diff
@@ -487,6 +487,159 @@ AlgoliaSearch.prototype.batch = function(operations, callback) {
...
+/**
+ * List all the clusters
+ *
+ * (the callback is the second argument)
+ *
+ * @return {Promise|undefined} Returns a promise if no callback given
+ * @example
+ * client.listClusters();
+ */
+AlgoliaSearch.prototype.listClusters = function(_, callback) {
```
vvo: Same, why _, callback?

src/AlgoliaSearch.js
```diff
@@ -487,6 +487,159 @@ AlgoliaSearch.prototype.batch = function(operations, callback) {
...
+ * @example
+ * client.listClusters();
+ * client.listClusters({ page: 3, hitsPerPage: 30});
+ */
+AlgoliaSearch.prototype.listUserIDs = function(data, callback) {
+  return this._jsonRequest({
+    method: 'GET',
+    url: '/1/clusters/mapping',
+    body: data,
+    hostType: 'write',
```
vvo: read?

src/AlgoliaSearch.js
```diff
@@ -487,6 +487,159 @@ AlgoliaSearch.prototype.batch = function(operations, callback) {
...
+ *
+ * @return {Promise|undefined} Returns a promise if no callback given
+ * @example
+ * client.listClusters();
+ */
+AlgoliaSearch.prototype.listClusters = function(_, callback) {
+  return this._jsonRequest({
+    method: 'GET',
+    url: '/1/clusters',
+    hostType: 'write',
```
vvo: read?

src/AlgoliaSearch.js
Haroenv: because this way there's no surprise if this method will later need options 🤔

src/AlgoliaSearch.js
Haroenv: What decides which is read and which is write?

src/AlgoliaSearch.js
vvo: Given the state of this JS client I think it's best to then do a non-breaking change evolution (from no params to params).

So we must support:
listClusters() => promise
listClusters(callback)

Calling listClusters(null, callback) only in case we add new features is too magic/uncommon

src/AlgoliaSearch.js
vvo: read is anything that is not modifying the index

PR #648, opened 2017-12-29
fix(core): make sure api key is sent as body if > 500 chars

### LINKED ISSUE #647: Search generating too long URLs for large API keys
__Note: This problem seems to be specific to the most recent release, 3.24.8. Less recent versions, including 3.24.7, don't seem to have this behavior. See below comment__

We've found what appears to be a bug with the current release version. In cases where an API key is considerably long, performing a search will include it in the URL and cause Algolia's DSN server to throw a 414.

This seems to be very similar to an issue that was patched before: https://github.com/algolia/algoliasearch-client-javascript/issues/319

While the JS client code seems to be doing a POST and putting our API key in the body, for some reason it is still including it as a query parameter as well, which is what causes the 414. Simply replaying the url and removing the API key from the query works fine (tested by copying as cURL command from Chrome's network tab).

It's also worth noting that this appears to have happened after release __3.22.1__. We were able to get this older version of your vended JS by requesting the un-minified version for a time (`<script src="//cdn.jsdelivr.net/algoliasearch/3/algoliasearch.js"></script>`) and confirmed it did not have the problem we are seeing. I took some screenshots of my Chrome dev tools to help show this.

__Version 3.22.1__

_Sources_

3-22-1-sources

_Network_

3-22-1-networktab1
3-22-1-networktab2

__Version 3.24.8__

_Sources_

3-24-8-sources

_Network_

3-24-8-networktab1
3-24-8-networktab2
3-24-8-networktab3

Haroenv:
This wasn't the case in the latest patch, because `withAPIKey` is not the same as `withApiKey`. I thought the bug was way worse before I selected the text and wondered why there were only two 'matches' in my editor, rather than three, which I knew was the case. Anyway, long story short, it was just a typo

fixes #647

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -488,7 +488,7 @@ AlgoliaSearchCore.prototype._getSearchParams = function(args, params) {
  * Compute the headers for a request
  *
  * @param [string] options.additionalUA semi-colon separated string with other user agents to add
- * @param [boolean=true] options.withAPIKey Send the api key as a header
+ * @param [boolean=true] options.withApiKey Send the api key as a header
  * @param [Object] options.headers Extra headers to send
  */
 AlgoliaSearchCore.prototype._computeRequestHeaders = function(options) {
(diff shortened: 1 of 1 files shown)


Haroenv:
Travis can't run because of #649, tests pass locally, I'm adding a test case for this (weird that there wasn't one yet 🤔)

PR #650, opened 2018-01-02
feat(getSettings): allow `advanced=1`

### LINKED ISSUE #637: Add request options to the getSettings method
The method `getSettings ` is missing a way to pass request options along with the callback

https://github.com/algolia/algoliasearch-client-javascript/blob/fa9506a81fb6ba065024b980d5f8ed17163132ad/src/Index.js#L604-L619

Because of that it's not possible to pass "advanced=1;

Haroenv:
cc @seafoox

**Summary**

There was no way to ask for `?advanced=1` in `getSettings`

**Result**

a first parameter to `getSettings` (with fallback to the callback) which is an object

if `advanced` is present in the object, it's added to the url

Fixes #637

This is an undocumented parameter, which has enterprise usage, which is why I didn't add it to the docblock.

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -608,11 +608,20 @@ Index.prototype.clearIndex = function(callback) {
 *  error: null or Error('message')
 *  content: the settings object or the error message if a failure occured
 */
-Index.prototype.getSettings = function(callback) {
-  var indexObj = this;
+Index.prototype.getSettings = function(opts, callback) {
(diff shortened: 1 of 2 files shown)


vvo:
> This is an undocumented parameter, which has enterprise usage, which is why I didn't add it to the docblock.

the docblock for now is now used for automatic generation but more as a dev help, I would still add it with the comment you put in the PR desc, that would help understand I believe.

PR #656, opened 2018-01-11
chore(test): use API Key dealer

Haroenv:
**Summary**

use api key dealer instead of hardcoded api keys so we can accept integration tests from non-owners of the repository

**Result**

no security flaw, but still integration tests


Haroenv:
locally I don't get a 403, but there I do 🤔 cc @julienbourdeau

PR #663, opened 2018-02-07
fix(exportXXX): bind the methods

### LINKED ISSUE #660: exportRules() throws ⚡️ `Cannot read property '_jsonRequest' of undefined`

bug

After loading https://cdn.jsdelivr.net/algoliasearch/3/algoliasearch.min.js (v3.24.9), the JavaScript code below throws `TypeError: Cannot read property '_jsonRequest' of undefined`

```
const fromClient = algoliasearch(fromAppId, fromApiKey);
const fromIndex = fromClient.initIndex(fromIdxName);
fromIndex.exportRules(100, console.log);
```

Tested on both nodejs (back-end) and in jsfiddle (front-end) with the same version.

I relied on the following documentation: https://www.algolia.com/doc/api-reference/api-methods/rules-export/?language=javascript#examples

I expected to have a list query rules printed to my javascript console.

Haroenv:
**Summary**

These two errors were unnoticed, because the integration tests apparently bind everything.

**Result**

fixes #660

No more error about `this.as._jsonRequest` being undefined

### DIFF
--- a/test/spec/common/index/exportRules.js
+++ b/test/spec/common/index/exportRules.js
@@ -1,0 +1,166 @@
+'use strict';
+
+var test = require('tape');
+var bind = require('lodash-compat/function/bind');
+var createFixture = require('../../../utils/create-fixture');
+var fauxJax = require('faux-jax');
+var parse = require('url-parse');
+var arrayFrom = require('array.from');
+
+test('exportRules()', function(t) {
+  t.plan(6);
+  var fixture = createFixture();
+  var index = fixture.index;
+  fauxJax.install({gzip: true});
+  fauxJax.once('request', firstBrowse);
+
+  index
+    .exportRules()
+    .then(function(data) {
+      t.ok(data, 'got some rules back');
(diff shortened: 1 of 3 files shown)


test/spec/common/index/exportRules.js
```diff
@@ -0,0 +1,166 @@
+'use strict';
```
vvo: Why not just a test-case/ https://github.com/algolia/algoliasearch-client-javascript/tree/develop/test/spec/common/client/test-cases?

Haroenv: Because it's not a single API call, it's more like `browseAll`

vvo: got it

vvo:
Can you also add a return here:

https://github.com/algolia/algoliasearch-client-javascript/blob/58a4b2dbe6258769a869e2668d0d66367359504c/src/Index.js#L671-L674

This should be callback(); return;

so that you do not return the promise too (nothing major but still)

PR #664, opened 2018-02-08
fix(getSettings): allow empty call

Haroenv:
**Summary**

in https://github.com/algolia/algoliasearch-client-javascript/commit/f91cddec5bbbb45b43aac15120723f99e5737608 accidentally a first argument was necessary.

**Result**

no more error with empty arguments

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -608,12 +608,15 @@ Index.prototype.clearIndex = function(callback) {
 * @param opts.advanced get more settings like nbShards (useful for Enterprise)
 * @param callback (optional) the result callback called with two arguments
 *  error: null or Error('message')
-*  content: the settings object or the error message if a failure occured
+*  content: the settings object or the error message if a failure occurred
 */
 Index.prototype.getSettings = function(opts, callback) {
-  if (arguments.length === 1 || typeof opts === 'function') {
-    callback = opts;
-    opts = {};
+  if (arguments.length < 2) {
+    if (typeof opts === 'function') {
+      callback = opts;
+      opts = {};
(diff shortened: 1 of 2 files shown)


test/run-integration.js
```diff
@@ -137,6 +138,20 @@ function setSettings(t) {
...
 
+function getSettings(t) {
+  t.plan(2);
+
+  index
+    .setSettings({attributesForFaceting: ['searchable(category)']})
+    .then(function() {return index.getSettings({advanced: 1});})
+    .then(get('attributesForFaceting'))
+    .then(_.partialRight(t.deepEqual, ['searchable(category)'], 'Settings were get (advanced)'))
+    .then(index.getSettings)
```
vvo: Here getSettings is called with one argument (result from previous Promise, even `undefined` wdyt?

Haroenv: good call, amending the test to be really with no arguments

src/Index.js
```diff
@@ -608,12 +608,15 @@ Index.prototype.clearIndex = function(callback) {
 * @param opts.advanced get more settings like nbShards (useful for Enterprise)
 * @param callback (optional) the result callback called with two arguments
 *  error: null or Error('message')
-*  content: the settings object or the error message if a failure occured
+*  content: the settings object or the error message if a failure occurred
 */
 Index.prototype.getSettings = function(opts, callback) {
-  if (arguments.length === 1 || typeof opts === 'function') {
```
vvo: You could also have added opts = opts || {} after the original if no? Would that have worked?

PR #669, opened 2018-02-28
fix(nodejs): handle network errors

vvo:
Before this commit we were not listening for http network errors
that can happen once the connection is opened. Those can happen when you
have a network issue in the middle of a response.

I was able to reproduce this error by launching multiple requests and
unplugging my network altogether.

I would constantly get the infamous "zlib error" which was only an
effect of not being able to continue the response listening.


Haroenv:
This shouldn't be an issue on v4 since we handle `onError` already. Thanks for fixing it here!

PR #670, opened 2018-03-05
feat(security): use https by default

vvo:
Originally the fact to use the page protocol or http by default was made
to speed up first requests to avoid initiating an https connection.

Nowadays this is no more of a big issue performance wise (citation
needed), but security wise, the use of http by default combined with
JSONP can be a big issue of MITM attacks.


Haroenv:
Releasing tomorrow

asurak:
Is it released?

Haroenv:
@asurak ready to release but just now had a doubt if this might not accidentally suppress the error when people run any of the methods we guard to only run on https

Haroenv:
Never mind, was confused. That error is in the server, and that's why I didn't find it in the code base 👍

Haroenv:
3.25.0

PR #671, opened 2018-03-07
fix(protocol): avoid throwing when protocol found

vvo:


Haroenv:
Just confirmed that this fixes the examples

hkdobrev:
Thanks for fixing this! However, I want to express my frustration with how this code was released. The Travis CI build is still broken and you release new versions of the library without having a green build. This is really prohibitive to a good testing culture. Looking at the code change, this would have been easily caught by a unit test.

The effective result for us was a lot of the JavaScript on our website didn't work even after the fix was released, because of the day cdn.js cache the content on the major version and we are bundling the code using the SDK with more JS with Browserify.

So indeed, we were affected, because we depend on just `/3/` from the CDN, but the cache invalidation there is not great when an immediate fix needs to be released.

vvo:
@hkdobrev Thanks for jumping in here,

> I want to express my frustration with how this code was released. The Travis CI build is still broken and you release new versions of the library without having a green build. This is really prohibitive to a good testing culture. Looking at the code change, this would have been easily caught by a unit test.

You are totally right and I apologise for the mess this created on your side. We have been pushing back updating our test suite setup for too long which have made us being less good at ensuring the stability of our JavaScript client.

Here are the steps we decided to take before any new release will happen on this client:
- remove the current set of tests that are using zuul
- replace them with single unit tests on node
- create a single good functional test with a real browser that will build the library and use it the way a user would use it (because unit tests are not always sufficient)
- remove any reference in our documentation and integration that is using /3/ because auto-updating is never a good idea. This is never the case on the backend, the fact that jsDelivr provides this is nice for POCs, not for production websites. You don't want your website to be updated by a third party.

> The effective result for us was a lot of the JavaScript on our website didn't work even after the fix was released, because of the day cdn.jsdelivr caches the content on the major version and we are bundling the code using the SDK with more JS with Browserify.

We just purged jsDelivr 30 minutes ago the right way so that /3/ is now delivering the right fix.

> So indeed, we were affected, because we depend on just /3/ from the CDN, but the cache invalidation there is not great when an immediate fix needs to be released.

Yes indeed the cache invalidation is tricky to handle well because of CDN cache and browser cache. Right now the version on jsDelivr is the right one though.

We advise you to move over to using a specific version. Since if your website is working on day one, and you don't need any new features, you can stay on a specific version. As we do as devs on backend dependencies, they are not auto update luckily.

Again sorry about this bad code change and release and thanks for the rapid feedback on it. We will ensure next releases are more robust.

hkdobrev:
The steps sound right and we already switched to a fixed version! Thanks for the update!

PR #676, opened 2018-03-20
feat(core): allow disabling of dsn

### LINKED ISSUE #675: Add an option to disable DSN

feature

If you have DSNs around the world, but are also doing backend search, some providers aren't able to resolve the correct DSN to route to, and thus round-robin all over the world.

**If the current behavior is a bug, please provide all the steps to reproduce and a minimal
[JSFiddle](https://jsfiddle.net/) example or a repository on GitHub that we can `npm install`
and `npm start`.**

You should be able to disable DSN in certain cases.

Proposed option for this:

```js
algoliasearch(appId, apiKey, { dsn: false });
```

This option will make sure that the first read host is `appId.algolia.net`, and not the default `appId-dsn.algolia.net`.

Haroenv:
**Summary**

In some cases dsn needs to be disabled, and the base url needs to be requested.

closes #675 

**Result**

A new option `dsn`, which defaults to true for when you want to disable it

```js
algoliasearch(appId, apiKey, {dsn: false});
```

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -94,7 +94,8 @@ function AlgoliaSearchCore(applicationID, apiKey, opts) {
     });
 
     // no hosts given, compute defaults
-    this.hosts.read = [this.applicationID + '-dsn.algolia.net'].concat(defaultHosts);
+    var mainSuffix = (opts.dsn === false ? '' : '-dsn') + '.algolia.net';
+    this.hosts.read = [this.applicationID + mainSuffix].concat(defaultHosts);
     this.hosts.write = [this.applicationID + '.algolia.net'].concat(defaultHosts);
   } else if (isArray(opts.hosts)) {
     // when passing custom hosts, we need to have a different host index if the number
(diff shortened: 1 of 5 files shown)


src/AlgoliaSearchCore.js
```diff
@@ -94,7 +94,8 @@ function AlgoliaSearchCore(applicationID, apiKey, opts) {
     });
 
     // no hosts given, compute defaults
-    this.hosts.read = [this.applicationID + '-dsn.algolia.net'].concat(defaultHosts);
+    var mainSuffix = (opts.dsn === false ? '' : '-dsn') + '.algolia.net';
```
Haroenv: tried to be as safe as possible here and only disable if it's explicitly `false`, but maybe another strategy works better @vvo

Haroenv:
Checked and the tests are only failing at the end of the Zuul bit, the actual tests are passing

src/AlgoliaSearchCore.js
vvo: opts is always defined in the code so this can't fail. This it the right way to do it.

PR #677, opened 2018-03-27
feat(sffv): Add `client.searchForFacetValues()` 

francoischalifour:
_This PR adds the top-level `client.searchForFacetValues([{ indexName, params }, ...queries])` method._

## Why

As we're going towards a more backend agnostic InstantSearch, we'd like to polish the internal JS client API to make it easier to write and understand.

With the current API, we need to initialize an index and to request SFFV on this index. This is an implementation detail that doesn't need to be exposed to our users.

Since AlgoliaSearch Transformers leverage this JS client interface to allow creation of custom backend implementations, we need to make it smoother. `searchForFacetValues()` should be at the same level as `search()`. For consistency, we also need to support multiple facet values queries (take an array of queries as parameter instead of a single query).

Right now, if you want to use a custom implementation backend that supports SFFV, you need to know the implementation:

```javascript
client.initIndex(indexName)
  .searchForFacetValues(params)
  .then(({ facetHits }) => {
      res.status(200).send(facetHits);
  });
```

## Proposal

Let's skip this `initIndex(indexName)` method and pass the `indexName` in the query object of the new `client.searchForFacetValues([{ indexName, params }, ...queries])` method.

The new usage will be easier and more powerful/flexible:

```javascript
client.searchForFacetValues([{ indexName, params }])
  .then(contents => {
    const [firstContent] = contents;
    const { facetHits } = firstContent;

    res.status(200).send(facetHits);
  });
```

## Backward compatibility

We should make our helper use this new method by default but fall back to the old one if the client used doesn't expose it. I will make a PR for this!

----

~_In this series of improvements, we'll make the helper use the promise version of `search(queries)` soon._~ ✅

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -610,6 +610,60 @@ AlgoliaSearchCore.prototype.search = function(queries, opts, callback) {
 };
 
 /**
+* Search for facet values
+* https://www.algolia.com/doc/rest-api/search#search-for-facet-values
+* This is the top-level API for SFFV.
+*
+* @param {string} query.indexName Index name, name of the index to search.
+* @param {object} query.params Query parameters.
+* @param {string} query.params.facetName Facet name, name of the attribute to search for values in.
+* Must be declared as a facet
+* @param {string} query.params.facetQuery Query for the facet search
+* @param {string} [query.params.*] Any search parameter of Algolia,
+* see https://www.algolia.com/doc/api-client/javascript/search#search-parameters
+* Pagination is not supported. The page and hitsPerPage parameters will be ignored.
+* @param callback (optional)
+*/
+AlgoliaSearchCore.prototype.searchForFacetValues = function(query, callback) {
+  var usage =
+    'Usage: client.searchForFacetValues({indexName, {facetName, facetQuery, ...params}}[, callback])';
+
+  if (
+    !query ||
+    query.indexName === undefined ||
+    query.params.facetName === undefined ||
+    query.params.facetQuery === undefined
+  ) {
+    throw new Error(usage);
+  }
+
+  var clone = require('./clone.js');
+  var omit = require('./omit.js');
+
+  var indexName = query.indexName;
+  var params = query.params;
+
+  var facetName = params.facetName;
+  var filteredParams = omit(clone(params), function(keyName) {
+    return keyName === 'facetName';
+  });
+  var searchParameters = this._getSearchParams(filteredParams, '');
+
+  return this._jsonRequest({
+    method: 'POST',
+    url:
+      '/1/indexes/' +
+      encodeURIComponent(indexName) +
+      '/facets/' +
+      encodeURIComponent(facetName) +
+      '/query',
+    hostType: 'read',
+    body: {params: searchParameters},
+    callback: callback
+  });
+};
+
+/**
  * Set the extra security tagFilters header
  * @param {string|array} tags The list of tags defining the current security filters
  */
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -636,7 +636,7 @@ AlgoliaSearchCore.prototype.searchForFacetValues = function(queries) {
 
   var client = this;
 
-  return map(queries, function performQuery(query) {
+  return Promise.all(map(queries, function performQuery(query) {
     if (
       !query ||
       query.indexName === undefined ||
@@ -670,7 +670,7 @@ AlgoliaSearchCore.prototype.searchForFacetValues = function(queries) {
       hostType: 'read',
       body: {params: searchParameters}
     });
-  });
+  }));
 };
 
 /**
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -614,52 +614,62 @@ AlgoliaSearchCore.prototype.search = function(queries, opts, callback) {
 * https://www.algolia.com/doc/rest-api/search#search-for-facet-values
 * This is the top-level API for SFFV.
 *
-* @param {string} query.indexName Index name, name of the index to search.
-* @param {object} query.params Query parameters.
-* @param {string} query.params.facetName Facet name, name of the attribute to search for values in.
+* @param {object[]} queries An array of queries to run.
+* @param {string} queries[].indexName Index name, name of the index to search.
+* @param {object} queries[].params Query parameters.
+* @param {string} queries[].params.facetName Facet name, name of the attribute to search for values in.
 * Must be declared as a facet
-* @param {string} query.params.facetQuery Query for the facet search
-* @param {string} [query.params.*] Any search parameter of Algolia,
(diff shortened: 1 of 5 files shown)


francoischalifour:
The Angular bundle size is [120B bigger than anticipated](https://bundlesize-store.now.sh/build?info=%7B%22files%22%3A%5B%7B%22maxSize%22%3A18432%2C%22path%22%3A%22.%2Fdist%2Falgoliasearch.angular.min.js%22%2C%22size%22%3A18265%7D%2C%7B%22maxSize%22%3A18432%2C%22path%22%3A%22.%2Fdist%2Falgoliasearch.jquery.min.js%22%2C%22size%22%3A18128%7D%2C%7B%22maxSize%22%3A47104%2C%22path%22%3A%22.%2Fdist%2Falgoliasearch.angular.js%22%2C%22size%22%3A47229%7D%2C%7B%22maxSize%22%3A47104%2C%22path%22%3A%22.%2Fdist%2Falgoliasearch.jquery.js%22%2C%22size%22%3A46810%7D%2C%7B%22maxSize%22%3A47104%2C%22path%22%3A%22.%2Fdist%2Falgoliasearch.parse.js%22%2C%22size%22%3A27745%7D%2C%7B%22maxSize%22%3A13312%2C%22path%22%3A%22.%2Fdist%2FalgoliasearchLite.min.js%22%2C%22size%22%3A12532%7D%2C%7B%22maxSize%22%3A32768%2C%22path%22%3A%22.%2Fdist%2FalgoliasearchLite.js%22%2C%22size%22%3A32377%7D%5D%2C%22repo%22%3A%22algolia%2Falgoliasearch-client-javascript%22%2C%22branch%22%3A%22feat%2Fsffv-top-level%22%2C%22commit_message%22%3A%22Merge%20994c742d67c1d43f73bebf599d08db239553b043%20into%20f7730e52bdd52e4e93706de3c57aa4a395f53a62%22%2C%22sha%22%3A%22994c742d67c1d43f73bebf599d08db239553b043%22%7D), which is why the build fails. We can increase the bundle size threshold from 46kB to 46.5kB if you're OK with that.

vvo:
Thanks!

Just to be clear, this is right now needed to make it easier for us internally to create fake clients from transformers, is that right?

Good way to also raise the fact that we have a dual API in all API clients where everything under index.* could maybe just be sitting in client.* with indexName being available for some of those methods as a parameter.

cc @julienbourdeau worth investigating while reviewing the APIs of APIs clients

francoischalifour:
> Just to be clear, this is right now needed to make it easier for us internally to create fake clients from transformers, is that right?

It is actually more than that. Let's say you want to implement Elasticsearch on a custom backend with InstantSearch like [this example](https://github.com/algolia/algoliasearch-transformers/blob/707e6982470d6ce3be3e639b718e93e31ad82797/packages/elasticsearch-transformer/examples/concerts-server/index.js#L58). You'd need to use `search()` and `initIndex.searchForFacetValues()` (instead of `searchForFacetValues()`). With this new API method, you could skip the "pointless" `initIndex()` call (from the user point of view).

To sum up, it would simplify the surface of the API that the user needs to know to use a custom backend implementation via a transformer.

Does that make sense?

Haroenv:
If you're not using Algolia, why would you use the Algolia client backend?

francoischalifour:
@Haroenv Are you referring to the `concert-server` example? This file has different endpoints to showcase all the possibilities. I'm not using `algoliasearch` for the Elasticsearch endpoints, but for the custom Algolia endpoint.

bobylito:
Obviously I'm in favor of that PR, since I discussed it with @francoischalifour beforehand.

After discussing it with @Haroenv who had some reserve about it, there is a slight problem. This feature pushes for simplicity and symmetry. However this does not have complete symmetry with search. It is not able to do a multi index SFFV.

And that sounds like an actual good idea since we have a use case for this: a smart search box that would be able to complete facet values over multiple facetted attributes. Implementing the multi index search for facet values would be an enabler of such new UI pattern.

vvo:
Agreed on implementing .searchForFacetValues([{indexName, ...}]).

About the example on https://github.com/algolia/algoliasearch-transformers/blob/707e6982470d6ce3be3e639b718e93e31ad82797/packages/elasticsearch-transformer/examples/concerts-server/index.js#L58, I understand now why it's easier to have searchForFacetValues at the root for such usecase where you really want to have the transformer acting on the backend and not on the frontend.

francoischalifour:
I'm going to update this PR to pass an array of queries to `searchForFacetValues()`. It turns out it's not an easy task because of the testing environment current limitations.

francoischalifour:
@Haroenv This is too painful to unit test this new feature because of the current way we deal with unit tests (our custom framework). I've tested the feature with a integration test and it works fine (@bobylito approved verbally).

Let me know what you think!

test/run-integration.js
```diff
@@ -164,6 +165,27 @@ function searchForFacetValues(t) {
     .then(noop, _.bind(t.error, t));
 }
 
+function searchForFacetValues(t) {
+  t.plan(2);
+
+  client
+    .searchForFacetValues([
+      {
```
Haroenv: would maybe be interesting to test one with a single query only too, because I imagine that will be used more

src/AlgoliaSearchCore.js
```diff
@@ -609,6 +609,70 @@ AlgoliaSearchCore.prototype.search = function(queries, opts, callback) {
...
+  }
+
+  var client = this;
+
+  return Promise.all(map(queries, function performQuery(query) {
+    if (
+      !query ||
+      query.indexName === undefined ||
+      query.params.facetName === undefined ||
+      query.params.facetQuery === undefined
```
samouss: I'm not following the all picture of this discussion but why do we nest the `facetName`, `facetQuery` and the rest of the parameters in a `params` object? The signature of the function will be a bit less verbose.

```js
client.searchForFacetValues([
  {
    indexName: "products",
    params: {
      facetName: "brand",
      facetValue: "Apple"
      ...searchParameters,
    }
  }
]);

// vs

client.searchForFacetValues([
  {
    indexName: "products",
    facetName: "brand",
    facetValue: "Apple",
    ...searchParameters,
  }
]);
```

Haroenv: agreed with the second syntax, we use this kind of pattern in other multi-index methods too

bobylito: That's a good point if it matches the other method signatures.

samouss: I didn't look correctly to the `search()` signature of the client. It looks the same as the one implemented in the PR.

https://github.com/algolia/algoliasearch-client-javascript/blob/2533c50a9588f5eccff1ebcf9716f476451d9f60/src/AlgoliaSearchCore.js#L564-L578

francoischalifour: It's the same signature as `search()` indeed. Although I agree `params` is not necessary.

test/run-integration.js
francoischalifour: Thanks, I will add the test!

PR #684, opened 2018-04-23
feat(exports): add errors to exported module

### LINKED ISSUE #683: Expose AlgoliaSearchError
I'd like to be able to conditionally check if I received an `AlgoliaSearchError`, using:
```js
const AlgoliaSearchError = ???;
try {
  await index.getSettings();
} catch (e) {
  if (e instanceof AlgoliaSearchError) {
    ...
  }
}
```
To my suprise, I realized that I couldn't find any reference to the class.
Could we expose it at the top level `module.exports` ?

Jerska:
**Summary**

Closes #683

Exposes all internal client errors.
I'm not sure if only `AlgoliaSearchError` can bubble up.
If this is the case, close this PR, and use this one instead: https://github.com/algolia/algoliasearch-client-javascript/pull/685

**Result**

```sh
$ node
> require('./index');
{ [Function: algoliasearch]
  errors:
   { AlgoliaSearchError: { [Function: AlgoliaSearchError] super_: [Object] },
     UnparsableJSON: { [Function: AlgoliaSearchCustomError] super_: [Object] },
     RequestTimeout: { [Function: AlgoliaSearchCustomError] super_: [Object] },
     Network: { [Function: AlgoliaSearchCustomError] super_: [Object] },
     JSONPScriptFail: { [Function: AlgoliaSearchCustomError] super_: [Object] },
     JSONPScriptError: { [Function: AlgoliaSearchCustomError] super_: [Object] },
     Unknown: { [Function: AlgoliaSearchCustomError] super_: [Object] } },
  version: '3.27.0',
  ua: 'Algolia for Node.js 3.27.0',
  initPlaces: [Function: places] }
```

### DIFF
--- a/src/browser/builds/algoliasearch.angular.js
+++ b/src/browser/builds/algoliasearch.angular.js
@@ -33,6 +33,7 @@ window.angular.module('algoliasearch', [])
       return new AlgoliaSearchAngular(applicationID, apiKey, opts);
     }
 
+    algoliasearch.errors = require('../../errors.js');
     algoliasearch.version = require('../../version.js');
     algoliasearch.ua = 'Algolia for AngularJS ' + algoliasearch.version;
     algoliasearch.initPlaces = places(algoliasearch);
--- a/src/browser/builds/algoliasearch.jquery.js
+++ b/src/browser/builds/algoliasearch.jquery.js
@@ -29,6 +29,7 @@ function algoliasearch(applicationID, apiKey, opts) {
   return new AlgoliaSearchJQuery(applicationID, apiKey, opts);
 }
 
+algoliasearch.errors = require('../../errors.js');
 algoliasearch.version = require('../../version.js');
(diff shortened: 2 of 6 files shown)


Haroenv:
seems like the history is wrong here :)

Jerska:
Indeed, just corrected this. :)

Haroenv:
Mostly in favour of this PR then, but importing modules is nicer than having an extra property on `algoliasearch`, what do you think about that?

Jerska:
If we have a standard way of importing the error module that is not going to be backwards-compatible in a minor patch, I'm all in for that. :)
It's just not going to be usable in the front-end if loaded through a CDN link, but otherwise, I'd be happy with this.

Haroenv:
As discussed in #683, let's leave the current situation as is, since `instanceof` breaks across realms too, and `.name` works with native errors, as well as our "subclassed" errors.

PR #685, opened 2018-04-23
feat(exports): add AlgoliaSearchError to exported module

### LINKED ISSUE #683: Expose AlgoliaSearchError

Jerska:
**Summary**

Closes #683

Exposes all internal client errors.
I'm not sure if only `AlgoliaSearchError` can bubble up.
If this is the case, use this PR, otherwise use this one instead: #684

**Result**

```sh
$ node
> require('./index');
{ [Function: algoliasearch]
  AlgoliaSearchError: { [Function: AlgoliaSearchError] super_: { [Function: Error] stackTraceLimit: 10 } },
  version: '3.27.0',
  ua: 'Algolia for Node.js 3.27.0',
  initPlaces: [Function: places] }
```

### DIFF
--- a/src/browser/builds/algoliasearch.angular.js
+++ b/src/browser/builds/algoliasearch.angular.js
@@ -33,6 +33,7 @@ window.angular.module('algoliasearch', [])
       return new AlgoliaSearchAngular(applicationID, apiKey, opts);
     }
 
+    algoliasearch.AlgoliaSearchError = require('../../errors.js').AlgoliaSearchError;
     algoliasearch.version = require('../../version.js');
     algoliasearch.ua = 'Algolia for AngularJS ' + algoliasearch.version;
     algoliasearch.initPlaces = places(algoliasearch);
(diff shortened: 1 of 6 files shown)


Haroenv:
tests seem to fail at a point they didn't before in the `set-cache` test 🤔

Jerska:
Tests seem to be failing on master too, and I don't see any reason for this change to have an impact on them.

Haroenv:
As discussed in #683, let's leave the current situation as is, since `instanceof` breaks across realms too, and `.name` works with native errors, as well as our "subclassed" errors.

PR #687, opened 2018-04-25
feat: Polyfill `Promise.all()`

francoischalifour:
## Why

As reported by @vvo, `AlgoliaSearchCore.prototype.searchForFacetValues` does not work in IE11 because we rely on the native [`Promise.all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) since #677.

## Solution

* Polyfill the `all()` method in each `client._promise` build
* Use `client._promise.all()` instead of `Promise.all()`

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -9,5 +9,11 @@ echo "Test"
 echo "Test: node"
 ./scripts/test-node
 
+#####
+# TODO: remove before merging the PR
+echo "Test: ci"
+./scripts/test-ci
+#####
+
 echo "Test: lint"
 ./scripts/lint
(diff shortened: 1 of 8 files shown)


francoischalifour:
I only trust Google and my intuition here, it's not enough.

Integration tests still work so there's for sure no regression. I was hoping we run the browser tests on `develop`. Do we have a way to run them on this branch though?

Haroenv:
You can comment out the bits that check for node in scripts/test and scripts/test-noe

francoischalifour:
Yay, all browser tests pass \o/

vvo:
Perfect, great work!

PR #689, opened 2018-04-30
feat: Deprecate API keys methods on Index in favor of Client ones

aseure:
**Summary**

As discussed with @maxiloc and @julienbourdeau, the API key methods are supposed to only be implemented at the client level, not at the index level anymore. This is due to the fact that per-index keys are not available via the dashboard and we do not want users to use them anymore.

**Result**

N/A

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -989,11 +989,11 @@ Index.prototype.setSettings = function(settings, opts, callback) {
 };
 
 /*
- @deprecated see index.listApiKeys
- */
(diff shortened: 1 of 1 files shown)


Haroenv:
We can do another pass later to remove tests :)

julienbourdeau:
Deprecated method are still part of the public API so I'm not sure about removing tests. If you do, you need to make sure it would get broken later. 👌

PR #690, opened 2018-04-30
test: Remove API keys-related tests from the Index instance

aseure:
**Summary**

Remove tests related to API keys methods of the Index instance (following @Haroenv review on https://github.com/algolia/algoliasearch-client-javascript/pull/689).

**Result**

Some tests removed for deprecated methods.


test/spec/common/index/interface.js
```diff
@@ -22,9 +22,6 @@ test('AlgoliaSearch index API spec', function(t) {
   var expectedMethods = [
     'addObject',
     'addObjects',
-    'addApiKey',
```
Haroenv: tests will fail if you remove it here, this is testing that everything which is present on the prototype is actually listed here. 

Not a particularly useful test, but still. 

What you can do for clarity is: 

```js
var deprecatedMethods = [ ... ];
var validMethods = [ ... ];
var expectedMethods = validMethods.concat(deprecatedMethods);
```

aseure: If this is the only tests we have, I think this PR is indeed not useful cc @julienbourdeau. Let's keep the tests intact until we actually remove the methods in a next major version.

WDYT? @julienbourdeau @Haroenv

Haroenv:
let's keep the tests there until removed. Easier to see that way too :)

PR #692, opened 2018-05-05
fix(env): don't assume `process` is defined

### LINKED ISSUE #691: Angular 6 Uncaught ReferenceError: process is not defined
**UPDATE FROM MAINTAINERS**

The working workaround for this issue for now is to add:

```js
(window as any).process = {
  env: { DEBUG: undefined },
};
```

**/UPDATE FROM MAINTAINERS**

to your polyfill.ts file in you Angular app

Bug
Angular does not provide a shim for "process" anymore (https://github.com/angular/angular-cli/issues/9827#issuecomment-369578814). 

Error:

```
Uncaught ReferenceError: process is not defined
    at Object../node_modules/algoliasearch/src/AlgoliaSearchCore.js (vendor.js:175349)
    at __webpack_require__ (runtime.js:82)
    at Object../node_modules/algoliasearch/src/AlgoliaSearch.js (vendor.js:174675)
    at __webpack_require__ (runtime.js:82)
    at Object../node_modules/algoliasearch/src/browser/builds/algoliasearch.js (vendor.js:177945)
    at __webpack_require__ (runtime.js:82)
...
```

**If the current behavior is a bug, please provide all the steps to reproduce and a minimal
[JSFiddle](https://jsfiddle.net/) example or a repository on GitHub that we can `npm install`
and `npm start`.**
Just using the library in an Angular 6 throws the error

To use algoliasearch library in an Angular 6 app

Haroenv:
**Summary**

Most environments (webpack, browserify) by default define `process` as a shim for it.

However, some environements don't supply it, so we should always guard against it not being defined

fixes #691

(should be tested in a real angular app first)

**Result**

guards around every use of `process`

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -10,7 +10,9 @@ var store = require('./store.js');
 // proxies limit)
 var MAX_API_KEY_LENGTH = 500;
 var RESET_APP_DATA_TIMER =
-  (process && process.env.RESET_APP_DATA_TIMER && parseInt(process.env.RESET_APP_DATA_TIMER, 10)) ||
+  (typeof process !== 'undefined' &&
+    process.env.RESET_APP_DATA_TIMER &&
+    parseInt(process.env.RESET_APP_DATA_TIMER, 10)) ||
   60 * 2 * 1000; // after 2 minutes reset to first host
 
 /*
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -10,7 +10,7 @@ var store = require('./store.js');
 // proxies limit)
 var MAX_API_KEY_LENGTH = 500;
 var RESET_APP_DATA_TIMER =
-  process.env.RESET_APP_DATA_TIMER && parseInt(process.env.RESET_APP_DATA_TIMER, 10) ||
+  (process && process.env.RESET_APP_DATA_TIMER && parseInt(process.env.RESET_APP_DATA_TIMER, 10)) ||
   60 * 2 * 1000; // after 2 minutes reset to first host
 
 /*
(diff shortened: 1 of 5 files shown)


Haroenv:
shoot, this somehow makes browserify add a polyfill for `process` in the bundle 😡

vvo:
Can we make it so that the fix is inside Angular InstantSearch? 
The one you proposed @Haroenv, can we ship it inside Angular InstantSearch?

```
(window as any).process = {
  env: { DEBUG: undefined },
};
```

I am trying to find "the good way" to fix this in the community, as using process.env is common I am not sure of the right way to rewrite it.

Haroenv:
That could be an option, I’ll try that out

iam4x:
I've tried to hotfix this into Angular InstantSearch directly with:

```js
// HOTFIX: define process
if (typeof window !== "undefined") {
  (window as any).process = { env: {} };
} else if (typeof global !== "undefined") {
  (global as any).process = { env: {} };
}
```

As entry point before anything else, it does not work :/

Haroenv:
Since the workaround works in user land and this is stalled on the fact that it adds an additional `process` polyfill in the bundles, I'll close this. We can revisit later, especially https://github.com/algolia/algoliasearch-client-javascript/pull/692#discussion_r188345457 can probably be done regardless

keichinger:
@Haroenv  can we maybe re-open this issue as it still breaks any Webpack web-only build for those people that don't want to ship extra polyfills or shims?

This fix would not only improve some broken builds right now but is also necessary if you want to support Webpack 5, as it'll get rid of node polyfills (see https://twitter.com/wSokra/status/1076126619146899462).

It seemed like a good idea for now but like a necessity for the future, IMO :)

PR #694, opened 2018-05-27
feat(cache): cache the requests instead of responses

samouss:
**Summary**

This PR update the strategy for caching the requests. With the current implementation we cache the results when we receive the response. It means that when we perform multiple exact same requests in parallel the client will trigger multiple network requests. 

This PR introduce a different approach, we cache the promise of the requests. Now when multiple exact same requests are triggered in parallel the same promise is reused for all the requests. Only one network call can be trigger for a request.

List of the changes:

- add the `responseText` property to each `_request` implementation. The `jsonRequest` method now return an object with `{ body, responseText }` to properly resolve the cached value. It brings consistency in the `jsonRequest` response, we can now safely use both of the attributes.
- update the `jsonRequest` method to cache the result of `doRequest` instead of caching the raw response
- **EDIT**: enable the caching of request behind an option `_useRequestCache`. By default this option is `false` in that case the behaviour is the same as today. The user can choose to enable this option with `_useRequestCache: true`.

Questions:

- I didn't found a way to test the different builds. The PR adds the `responseText` to each of the implementation but I don't know how to test that the `_request` method correctly return the response.
- Some implementation might rely on the `cache` attribute of the client. We don't document it but it's still accessible on the client instance. The type of the value cached change from string to a promise. ~~To avoid a BC we can introduce an option for handle different caching strategy with the default value on the previous implementation.~~
- We do clear the cache in case of error now. ~Last but not least since we cache the results of `doRequest` we also cache the error when the retry strategy fails. It means that on subsequent requests the error will be fetch from the cache instead of doing a new call. We can avoid this issue by clearing the cache in case of error, but it's **not implemented** in this PR.~

**Before**

before

**After**

after

### DIFF
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -454,7 +454,12 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
   if (client._useCache && cache && cache[cacheID] !== undefined) {
     requestDebug('serving response from cache');
 
-    var promiseForCache = cache[cacheID];
+    var maybePromiseForCache = cache[cacheID];
+
+    // In case the cache is warmup with value that is not a promise
+    var promiseForCache = typeof maybePromiseForCache.then !== 'function'
+      ? Promise.resolve({responseText: maybePromiseForCache})
+      : maybePromiseForCache;
 
     return interopCallbackReturn(promiseForCache, function(content) {
       // In case of the cache request, return the original value
--- a/src/AlgoliaSearchCore.js
+++ b/src/AlgoliaSearchCore.js
@@ -184,6 +184,7 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
   var requestDebug = require('debug')('algoliasearch:' + initialOpts.url);
 
   var body;
+  var cacheID;
   var additionalUA = initialOpts.additionalUA || '';
   var cache = initialOpts.cache;
   var client = this;
@@ -222,23 +223,6 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
     client._checkAppIdData();
 
     var startTime = new Date();
-    var cacheID;
-
-    if (client._useCache) {
-      cacheID = initialOpts.url;
-    }
-
-    // as we sometime use POST requests to pass parameters (like query='aa'),
-    // the cacheID must also include the body to be different between calls
-    if (client._useCache && body) {
-      cacheID += '_body_' + reqOpts.body;
-    }
-
-    // handle cache existence
-    if (client._useCache && cache && cache[cacheID] !== undefined) {
-      requestDebug('serving response from cache');
-      return client._promise.resolve(JSON.parse(cache[cacheID]));
-    }
 
     // if we reached max tries
     if (tries >= client.hosts[initialOpts.hostType].length) {
@@ -341,11 +325,10 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
       });
 
       if (httpResponseOk) {
-        if (client._useCache && cache) {
-          cache[cacheID] = httpResponse.responseText;
-        }
-
-        return httpResponse.body;
+        return {
+          responseText: httpResponse.responseText,
+          body: httpResponse.body
+        };
       }
 
       var shouldRetry = Math.floor(status / 100) !== 4;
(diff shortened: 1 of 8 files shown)


src/AlgoliaSearchCore.js
```diff
@@ -438,6 +421,52 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
...
+        exitPromise(function() {
+          initialOpts.callback(null, callback(content));
+        }, client._setTimeout || setTimeout);
+      }, function nookCb(err) {
+        exitPromise(function() {
+          initialOpts.callback(err);
+        }, client._setTimeout || setTimeout);
+      });
+    } else {
+      return promise.then(function(content) {
```
bobylito: Why isn't that `promise.then(callback)`?

samouss: For no particular reason, I'll update the call to the function. I'll also rename the parameters of the function `interopCallbackReturn` to be a bit more explicit.

src/AlgoliaSearchCore.js
```diff
@@ -421,7 +421,18 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
...
 
+  function isCacheValidWithCurrentID() {
+    return client._useCache && cache && cache[cacheID] !== undefined;
+  }
+
   function interopCallbackReturn(promise, callback) {
+    promise.catch(function() {
+      // Release the cache on error
+      if (isCacheValidWithCurrentID()) {
+        delete cache[cacheID];
```
samouss: I use delete here, but in case its too slow can assign `undefined`.

src/AlgoliaSearchCore.js
```diff
@@ -438,7 +421,62 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
     }
   }
 
-  var promise = doRequest(
+  function isCacheValidWithCurrentID() {
+    return client._useCache && cache && cache[cacheID] !== undefined;
```
Haroenv: why did you make `cacheID` global instead of using it as a function parameter?

src/AlgoliaSearchCore.js
```diff
@@ -438,7 +421,62 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
...
+  function interopCallbackReturn(request, callback) {
+    request.catch(function() {
+      // Release the cache on error
+      if (isCacheValidWithCurrentID()) {
+        delete cache[cacheID];
+      }
+    });
+
+    // either we have a callback
+    // either we are using promises
```
Haroenv: move the second comment to the else?

Haroenv:
>Some implementation might rely on the cache attribute of the client. We don't document it but it's still accessible on the client instance. The type of the value cached change from string to a promise. To avoid a BC we can introduce an option for handle different caching strategy with the default value on the previous implementation.

What if we threw a warning on accessing the `.cache` outside of the client? Should be possible right? 

Still of the opinion that I can't see a use case of anyone directly using `.cache` so it isn't really that important that the implementation changes

src/AlgoliaSearchCore.js
samouss: I tried to stick at the pattern used in the code actually, I really don't like relying on global variable. I can change it but I might have some issues to name the variable since ESLint won't allow me to use the same the variable name (I think).

Haroenv:
```js
function isCacheValidWithCurrentID(currentCacheID) {
}
```

this would work right?

samouss: Yes it would without the destructuring part.

vvo:
> I didn't found a way to test the different builds. The PR adds the responseText to each of the implementation but I don't know how to test that the _request method correctly return the response.

What we could do is create a new unit test where you target the local server that is running and there we would do checks that when launching two searches at the same time, only one is received by the server.

vvo:
> Some implementation might rely on the cache attribute of the client. We don't document it but it's still accessible on the client instance. The type of the value cached change from string to a promise. To avoid a BC we can introduce an option for handle different caching strategy with the default value on the previous implementation.

Yes for the time being we could add an option to activate the new behaviour. It would allow us to be compatible, then next major we remove the old cache.

vvo:
> We do clear the cache in case of error now.

Can you precise what you mean?

src/AlgoliaSearchCore.js
```diff
@@ -438,7 +421,62 @@ AlgoliaSearchCore.prototype._jsonRequest = function(initialOpts) {
...
+  if (client._useCache && body) {
+    cacheID += '_body_' + body;
+  }
+
+  if (isCacheValidWithCurrentID()) {
+    requestDebug('serving response from cache');
+
+    var maybePromiseForCache = cache[cacheID];
+
+    // In case the cache is warmup with value that is not a promise
```
vvo: When does that happens?

vvo:
This is a nice addition, let's make it as a feature flag and release!

vvo:
Tests on develop were fixed. Now all unit and integration tests are running on every platform.

Haroenv:
just rebased to have passing tests

src/AlgoliaSearchCore.js
Haroenv: when a user passes `algoliasearch('', '', { cache: myWarmedUpCacheWhichIsObjectsNotPromise });` I think

samouss:
I've updated the PR and the description.

vvo:
> Some implementation might rely on the cache attribute of the client. We don't document it but it's still accessible on the client instance. The type of the value cached change from string to a promise

This is no more true right since you made it a feature flag?

vvo:
> I didn't found a way to test the different builds. The PR adds the responseText to each of the implementation but I don't know how to test that the _request method correctly return the response.

This is now tested via a functional/black box test I believe? If so we are good to go

samouss:
> This is no more true right since you made it a feature flag?

Yep indeed when the feature is not enable the type of the cache remains the same. It changes only when we enable the option.

> This is now tested via a functional/black box test I believe? If so we are good to go

Almost all the tests use the `_jsonRequest` function, since this function use the `_request` implement we should be good. What is was looking for is a unit test for each build.

vvo:
Let's goooo!

eglinetwork:
Please add the option to the defintions of Typescript. 

TS2345: Argument of type '{ _useRequestCache: boolean; }' is not assignable to parameter of type 'ClientOptions | undefined'.

Haroenv:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26904

note for the future @eglinetwork, anyone can do a PR to DefinitelyTyped. Currently TypeScript isn't an officially supported language for us, so we won't proactively update the typings unless for breaking changes, or things we are actively deprecating. Thanks for mentioning this thogh

PR #695, opened 2018-05-28
test(integration): skip specific test for iOS 11

Haroenv:
see failure in Travis: https://travis-ci.org/algolia/algoliasearch-client-javascript/builds/374300255#L3665

This test can be skipped since really it's no problem if people get their results faster than expected.

I left the second timing test in there, since we still want the second request not to timeout

(this test is not run on PRs, so you won't really be able to see the difference in a PR)


vvo:
(reviewed with you, agreed)

PR #701, opened 2018-05-31
test(unit): fix and reactivate unit tests in PRs

vvo:
Mostly due to the change on https:, let's reactivate them in PRs
otherwise we don't keep track of errors


src/AlgoliaSearch.js
```diff
@@ -507,7 +507,7 @@ AlgoliaSearch.prototype.assignUserID = function(data, callback) {
     body: {cluster: data.cluster},
     callback: callback,
     headers: {
-      'X-Algolia-User-ID': data.userID
+      'x-algolia-user-id': data.userID
```
vvo: there's no casing in HTTP headers per spec

test/spec/browser/request-strategy/JSONP-fallback-per-request.js
```diff
@@ -19,7 +19,8 @@ test('Request strategy uses JSONP fallback on a per requests basis', function(t)
         currentURL.host,
         currentURL.host
       ],
-      timeout: requestTimeout
+      timeout: requestTimeout,
+      protocol: 'http:'
```
vvo: when calling our mock server, we need to stick with http since it's only running in http. It does not change the test added value

Haroenv:
seems like it errors with "tunnel not found" when a browser retries (maybe because we don't exit correctly? https://travis-ci.org/algolia/algoliasearch-client-javascript/builds/386191557#L5039

Haroenv:
Tests __almost__ pass, but stuff seems to timeout still:

```
/home/travis/build/algolia/algoliasearch-client-javascript/node_modules/zuul/bin/zuul:332
            throw err.message;
            ^
safari@10: Did not receive any new output from browser for 180s, shutting down
error Command failed with exit code 1.
```

EDIT: might have just fixed it with a wrong `t.plan`

vvo:
image

vvo:
giphy 3

PR #702, opened 2018-06-11
fix(parse): correct path to debug module

vvo:
fixes parse build, weirdly webpack is not throwing when importing an
unkown module using target node


vvo:
Fixing but seems like Parse now supports regular Node.js modules, if that's the case I will just remove the specific parse code and building.

Waiting for a support answer on this.

vvo:
Still a valid fix as of now, will re-ask users

PR #705, opened 2018-06-13
chore(parse): remove parse build and examples

vvo:
There's no more the need to have a specific Parse build since parse can
now use any npm module. I also updated the Algolia documentation.

Our previous documentation was asking people to copy paste, and since
they were not able to require() the parse cloud either, there's no
breaking change and this will be transparent


Haroenv:
bundlesize broke in master too after the merge, not exactly sure why 🤔

PR #707, opened 2018-06-14
docs(do): more relevant error message on bad objectID type

anonkey:
**Summary**
When sending a bad objectID type, the error message is not very clear, e.g. if you send a Buffer instead of String by mistake it make you think that you send a null objectID whereas this is not the case.

**Result**

If sending a null objectID error message is 'Cannot delete an object without an objectID' and if sending a bad type objectID message is  'ObjectID must be a string'

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -223,7 +223,11 @@ Index.prototype.saveObjects = function(objects, callback) {
 */
 Index.prototype.deleteObject = function(objectID, callback) {
   if (typeof objectID === 'function' || typeof objectID !== 'string' && typeof objectID !== 'number') {
-    var err = new errors.AlgoliaSearchError('Cannot delete an object without an objectID');
+    var err = new errors.AlgoliaSearchError(
+      objectID && typeof objectID !== 'function'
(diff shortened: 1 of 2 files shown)


Haroenv:
Thanks! I think the integration tests will also need to be updated with this change though

anonkey:
I'm a little busy ATM i'll do this as soon as possible.

Haroenv:
Thanks, I'll take a look at writing them some time soon as well.

vvo:
This integration tests are failing because external PRS are not getting the right API keys for integration tests. I will push to the branch and see how it goes.

vvo:
was merged

vvo:
I am sorry, your commit authorship was removed by doing another PR, now I know how to handle those (for us maintainers):
- when a PR is opened by someone from the outside
- checkout PR to a local branch
- push it to origin
- open a new PR, this will trigger travis build for commit SHA
- close new PR once tests are green, merge the original one

anonkey:
It's okay lot of thanks

PR #708, opened 2018-06-18
feat(analytics): Introduce new Analytics object

julienbourdeau:
**Summary**

This pullrequest introduces AB Testing to the client. A new object `Analytics` was introduce to query the analytics APi.

See more details here: https://github.com/algolia/algoliasearch-client-php/pull/408

**NOTE**: Tests on analytics should be improved one day

### DIFF
--- a/src/browser/createAlgoliasearch.js
+++ b/src/browser/createAlgoliasearch.js
@@ -87,6 +87,17 @@ module.exports = function createAlgoliasearch(AlgoliaSearch, uaSuffix) {
       // breaks it and set it to false by default
       if (req instanceof XMLHttpRequest) {
         req.open(opts.method, url, true);
+
+        // note: this breaks everything else but makes the browser tests pass
+        // the analytics endpoint _needs_ it via header, and doesn't accept QS
+        req.setRequestHeader(
+          'x-algolia-application-id',
+          opts.headers['x-algolia-application-id']
+        );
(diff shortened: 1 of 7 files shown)


Haroenv:
@julienbourdeau is it possible that the analytics API doesn't accept API key / appID from the query string, and only as header?

src/browser/createAlgoliasearch.js
```diff
@@ -87,6 +87,17 @@ module.exports = function createAlgoliasearch(AlgoliaSearch, uaSuffix) {
...
+        // note: this breaks everything else but makes the browser tests pass
+        // the analytics endpoint _needs_ it via header, and doesn't accept QS
+        req.setRequestHeader(
+          'x-algolia-application-id',
+          opts.headers['x-algolia-application-id']
+        );
+        req.setRequestHeader(
+          'x-algolia-api-key',
+          opts.headers['x-algolia-api-key']
+        );
```
Haroenv: I added this manually as request header for analytics. I guess it needs to be an option since without setting them explicitly as header (default is query string) I get a 401 error by the analytics API.

This either needs to be an option, either `inlineHeaders` needs to be an option, either the analytics API would accept API key / app id to be sent via query string. 

Last option is my preference

julienbourdeau:
🎉 🎉  Thank you @Haroenv and @vvo 🙏

Issue #610, opened 2017-10-13
Deprecate similarSearch -> use regular search with similarQuery


Haroenv:
https://github.com/algolia/algoliasearch-client-javascript/blob/f3fbc1950c3da589b570daff8082e0ded190418c/src/IndexCore.js#L135

- [ ] deprecate
- [x] remove in v4
- [x] wiki entry

PR #727, opened 2018-10-30
chore(core): deprecate similarSearch

Haroenv:
fixes https://github.com/algolia/algoliasearch-client-javascript/issues/610

This is documented nowhere and is beta, has an alternative with "similarQuery" search param

**Summary**

similarSearch isn't necessary

wiki link is: https://github.com/algolia/algoliasearch-client-javascript/wiki/Deprecated#indexsimilarsearchquery-callback

**Result**

the method is now deprecated, so it can be removed more safely


Haroenv:
I need to look into why the integration tests in the browsers fail again, it's not related to this PR though

PR #731, opened 2018-11-20
Add copy* and replaceAll* methods

julienbourdeau:
**Summary**

The new PHP api client will introduce new method to easily replace all objects/synonyms/rules and copy settings/synonyms/rules. We decided to back port these helpers to current API client version.

**Result**

I'm working on the tests for `replaceAllObjects`, the other methods are basically aliasing already existing methods.

### DIFF
--- a/src/AlgoliaSearch.js
+++ b/src/AlgoliaSearch.js
@@ -85,6 +85,18 @@ AlgoliaSearch.prototype.copyIndex = function(srcIndexName, dstIndexName, scopeOr
   });
 };
 
+AlgoliaSearch.prototype.copySettings = function(srcIndexName, dstIndexName, callback) {
+  return this.copyIndex(srcIndexName, dstIndexName, ['settings'], callback);
+};
+
+AlgoliaSearch.prototype.copySynonyms = function(srcIndexName, dstIndexName, callback) {
+  return this.copyIndex(srcIndexName, dstIndexName, ['synonyms'], callback);
+};
+
+AlgoliaSearch.prototype.copyRules = function(srcIndexName, dstIndexName, callback) {
+  return this.copyIndex(srcIndexName, dstIndexName, ['rules'], callback);
+};
+
 /**
  * Return last log entries.
  * @param offset Specify the first entry to retrieve (0-based, 0 is the most recent log entry).
--- a/src/Index.js
+++ b/src/Index.js
@@ -783,6 +783,18 @@ Index.prototype.batchSynonyms = function(synonyms, opts, callback) {
   });
 };
 
+Index.prototype.replaceAllSynonyms = function(synonyms, opts, callback) {
+  if (typeof opts === 'function') {
+    callback = opts;
+    opts = {};
+  } else if (opts === undefined) {
+    opts = {};
+  }
+  opts.replaceExistingSynonyms = true;
+
+  return this.batchSynonyms(synonyms, opts, callback);
+};
+
 Index.prototype.searchRules = function(params, callback) {
   if (typeof params === 'function') {
     callback = params;
@@ -899,6 +911,18 @@ Index.prototype.batchRules = function(rules, opts, callback) {
     body: rules,
     callback: callback
   });
+};
+
+Index.prototype.replaceAllRules = function(rules, opts, callback) {
+  if (typeof opts === 'function') {
+    callback = opts;
+    opts = {};
+  } else if (opts === undefined) {
+    opts = {};
+  }
+  opts.clearExistingRules = true;
+
+  return this.batchSynonyms(rules, opts, callback);
 };
 
 /*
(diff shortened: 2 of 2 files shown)


src/Index.js
```diff
@@ -213,6 +213,35 @@ Index.prototype.saveObjects = function(objects, callback) {
...
+  var saveResponse = this.saveObjects(objects);
+  if (safe) {
+    this.waitTask(saveResponse.taskID);
+  }
+
+  var moveResponse = this.as.moveIndex(tmpIndexName, this.indexName);
+  if (safe) {
+    this.waitTask(moveResponse.taskID);s
+  }
+
```
julienbourdeau: Should I call the `callback` here 🤔

src/AlgoliaSearch.js
```diff
@@ -85,6 +85,18 @@ AlgoliaSearch.prototype.copyIndex = function(srcIndexName, dstIndexName, scopeOr
   });
 };
 
+AlgoliaSearch.prototype.copySettings = function(srcIndexName, dstIndexName, callback) {
```
nunomaduro: It's not defined on the CTS yet, but IMO we should add tests on those methods as-well.

src/Index.js
```diff
@@ -213,6 +213,35 @@ Index.prototype.saveObjects = function(objects, callback) {
...
+  var safe = opts.safe || false;
+  var tmpIndexName = this.indexName + '_' + Math.random() + '_tmp';
+
+  var copyResponse = this.as.copyIndex(this.indexName, tmpIndexName, ['settings', 'rules', 'synonyms']);
+  if (safe) {
+    this.waitTask(copyResponse.taskID);
+  }
+
+  var saveResponse = this.saveObjects(objects);
```
nunomaduro: The save objects doesn't perform a chunk `yet`. So you may need to perform a chunk here.

src/Index.js
```diff
@@ -213,6 +213,35 @@ Index.prototype.saveObjects = function(objects, callback) {
...
+  if (safe) {
+    this.waitTask(saveResponse.taskID);
+  }
+
+  var moveResponse = this.as.moveIndex(tmpIndexName, this.indexName);
+  if (safe) {
+    this.waitTask(moveResponse.taskID);s
```
nunomaduro: An extra `s` in this line.

src/Index.js
```diff
@@ -783,6 +812,18 @@ Index.prototype.batchSynonyms = function(synonyms, opts, callback) {
   });
 };
 
+Index.prototype.replaceAllSynonyms = function(synonyms, opts, callback) {
```
nunomaduro: We probably should discuss if we should or not verify the `objectID` of the provided synonyms.

src/Index.js
```diff
@@ -901,6 +942,18 @@ Index.prototype.batchRules = function(rules, opts, callback) {
   });
 };
 
+Index.prototype.replaceAllRules = function(rules, opts, callback) {
```
nunomaduro: We probably should discuss if we should or not verify the `objectID` of the provided rules.

src/Index.js
nunomaduro: I guess 😅

src/Index.js
```diff
@@ -213,6 +213,35 @@ Index.prototype.saveObjects = function(objects, callback) {
...
+    callback = opts;
+    opts = {};
+  } else if (opts === undefined) {
+    opts = {};
+  }
+
+  var safe = opts.safe || false;
+  var tmpIndexName = this.indexName + '_' + Math.random() + '_tmp';
+
+  var copyResponse = this.as.copyIndex(this.indexName, tmpIndexName, ['settings', 'rules', 'synonyms']);
```
vvo: This is not the way promises works unfortunately. Have a look at the old deleteBy implementation which is similar to what you want to accomplish: a meta method that calls other methods: https://github.com/algolia/algoliasearch-client-javascript/blob/develop/src/Index.js#L297

src/Index.js
```diff
@@ -213,6 +213,35 @@ Index.prototype.saveObjects = function(objects, callback) {
...
+  if (safe) {
+    this.waitTask(saveResponse.taskID);
+  }
+
+  var moveResponse = this.as.moveIndex(tmpIndexName, this.indexName);
+  if (safe) {
+    this.waitTask(moveResponse.taskID);
+  }
+
+  return [copyResponse, saveResponse, moveResponse];
```
vvo: I don't get this line or what it's doing?? What the responsibility of replaceAllObjects exactly?

src/Index.js
julienbourdeau: I don't think we should, engine will return an error if no objectID is set.

src/Index.js
julienbourdeau: My stupid command ⌘ key doesn't work well, it happens all the time 😭

src/Index.js
julienbourdeau: I will dig into this, thank you :)

PR #732, opened 2018-11-24
feat(places-reverse): expose reverse endpoint for places

JonathanMontane:
**Summary**

This PR exposes the reverse geocoding endpoint of Places in the JS client through a `.reverse(options, callback)` method. This would allow the places.js library to leverage the power of js client wrt to retry strategy and caching.

**Result**

Usage
```js
const places = algoliasearch.initPlaces(appId, apiKey);

places.reverse({
  aroundLatLng: '48.880397,2.32699',
  language: 'en',
  getRankingInfo: true,
  hitsPerPage: 20
});
```

Response

image

**Notes:**
- the way the reverse API accepts arguments is slightly different from a standard endpoint and will fail if passed a body with params. The parameters have to be passed as search parameters in the url. Which is unfortunate because I could have used `buildSearchMethod('reverse', '/1/places/reverse')` if it was standard.
- I wasn't sure how recent the syntax could be so I went for ES3 (use of `concat`). If you have a more recent syntax that would improve readability, or if I should update my code for an older target, please let me know.

### DIFF
--- a/test/run-integration.js
+++ b/test/run-integration.js
@@ -104,7 +104,7 @@ if (!isABrowser) {
 
 function initPlaces(placesAppId, placesApiKey) {
   return function(t) {
-    t.plan(3);
+    t.plan(2);
     var places = algoliasearch.initPlaces(placesAppId, placesApiKey);
 
     places.search('paris').then(function(res) {


test/run-integration.js
```diff
@@ -112,6 +112,12 @@ function initPlaces(placesAppId, placesApiKey) {
     }, function(e) {
       t.fail(e);
     });
+
+    places.reverse('48.880397, 2.326991').then(function(res) {
+      t.ok(res.nbHits, 'We got some results by querying `48.880397, 2.326991`');
```
Haroenv: shouldn't this be > 0 or something?

francoischalifour: `nbHits` is `0` when no hits – the intent would indeed be clearer with the expression `nbHits > 0`.

src/places.js
```diff
@@ -2,6 +2,19 @@ module.exports = createPlacesClient;
 
 var buildSearchMethod = require('./buildSearchMethod.js');
 
+function urlEncodeOptions(options) {
```
francoischalifour: Not sure if relevant but we already use [`querystring-es3`](https://github.com/mike-spainhower/querystring) in the project which should be able to replace this custom function.

JonathanMontane: Thanks for the info! I updated the code to use the library instead. 👍

test/run-integration.js
```diff
@@ -104,7 +104,7 @@ if (!isABrowser) {
 
 function initPlaces(placesAppId, placesApiKey) {
   return function(t) {
-    t.plan(1);
+    t.plan(2);
```
JonathanMontane: I am unclear on what `t.plan` do. Can anyone explain the purpose of that method for me ? Thanks!

Haroenv: it is to tell the test runner how many assertions will be done in this test, so that it knows when to bail in asynchronous contexts

src/places.js
```diff
@@ -24,6 +25,20 @@ function createPlacesClient(algoliasearch) {
     var client = algoliasearch(appID, apiKey, opts);
     var index = client.initIndex('places');
     index.search = buildSearchMethod('query', '/1/places/query');
+    index.reverse = function(aroundLatLng, options, callback) {
+      var encodedOptions = qs3.encode(options);
+      var encoded = ['aroundLatLng=' + encodeURIComponent(aroundLatLng)]
```
Haroenv: why is `aroundLatLng` not one of the options instead? This code would be simpler like that

JonathanMontane: I wanted to have an interface that was similar to the one we have for search e.g. `search(query, options, cb)`, with `aroundLatLng` being the "query" part for a reverse geocoding query but I agree that it would indeed be much simpler to have everything in the options field instead.

Will update 👍

JonathanMontane: I changed the method interface to support both `reverse(aroundLatLng: string)` and `reverse(options: object)`.

PR #733, opened 2018-11-28
Remove dependency isarray

styfle:
**Summary**

There is no more need for `isarray`

**Result**

No logical change...just drop support for IE8

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "foreach": "^2.0.5",
     "global": "^4.3.2",
     "inherits": "^2.0.1",
-    "isarray": "^2.0.1",
     "load-script": "^1.0.0",
     "object-keys": "^1.0.11",
     "querystring-es3": "^0.2.1",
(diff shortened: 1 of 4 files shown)


Haroenv:
unfortunately we _do_ support IE8 in this client, and thus can't remove the dependency directly. The dependency is tiny, so this isn't the biggest issue probably, however if you do feel like contributing here, an option would be to inline with clear comments that this is for IE8. What do you think?

Haroenv:
Thanks for catching that, the usage of IE8 is really low, and we'd like to remove it in a major version altogether. I'll leave this PR open, for when we will take a good look and refactor a lot of this code incrementally (next year likely), this single change isn't exactly "worth" a major version, but there's more things that are removable (like es6-promise in user land etc.)

Haroenv:
We will solve this when we do a breaking change individually, but thanks for the PR!

PR #734, opened 2018-12-03
ci: use newer version of static-module

Haroenv:
**Summary**

**Result**

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,6 +133,9 @@
     "xhr": "^2.4.0",
     "zuul": "^3.12.0"
   },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


Haroenv:
it fixes the integration test, but browser unit test still seems to have a mysterious error, cc @vvo

vvo:
@Haroenv nope no idea what's going on here. You should try to run it locally using Node.js 11.3.0 to debug it properly if you want. Zuul etc are very old stacks so making them work on Node.js 11.3.0 was never really planned eh!

vvo:
https://github.com/substack/bulkify/issues/18

PR #735, opened 2018-12-13
feat(personalization): add basic methods

tkrugg:
🏁 This PR is ready for review

Adding basic methods for POST and GET verbs.
I called them (set|get)Strategy because that's the name other API client seem to have adopted. (cf [php client PR](https://github.com/algolia/algoliasearch-client-php/pull/493/files#r241372167)).

The docs of the REST API is here:
https://www.algolia.com/doc/rest-api/personalization/?language=javascript#personalization


src/AlgoliaSearch.js
```diff
@@ -653,6 +653,52 @@ AlgoliaSearch.prototype.searchUserIDs = function(data, callback) {
...
+ *      "Add to cart": { score: 50, type: "conversion" },
+ *      Purchase: { score: 100, type: "conversion" }
+ *   },
+ *   facetsScoring: {
+ *      brand: { score: 100 },
+ *      categories: { score: 10 }
+ *   }
+ * });
+ */
+AlgoliaSearch.prototype.setStrategy = function(data, callback) {
```
tkrugg: will sync with API clients teams on a better naming for this function.

src/AlgoliaSearch.js
```diff
@@ -653,6 +653,52 @@ AlgoliaSearch.prototype.searchUserIDs = function(data, callback) {
...
+
+/**
+ * Get strategy for personalization
+ *
+ * @return {Promise|undefined} Returns a promise if no callback given
+ * @example
+ * client.getStrategy();
+ */
+
+AlgoliaSearch.prototype.getStrategy = function(callback) {
```
tkrugg: will sync with API clients teams on a better naming for this function.

src/AlgoliaSearch.js
```diff
@@ -653,6 +653,52 @@ AlgoliaSearch.prototype.searchUserIDs = function(data, callback) {
   });
 };
 
+/**
+ * Set strategy for personalization
+ *
+ * @param {Object} data
+ * @param {Object} data.eventsScoring Scoring to apply to events
+ * @param {Object} data.facetsScoring Scoring to apply on facets
```
tkrugg: I would have liked further documenting the content of theses objects but I don't see a pattern for doing so in the rest of the codebase.
Any inputs @Haroenv ?

Haroenv: is eventsScoring an object? please put the shape in the documentation, and not just a repeat of the name :)

tkrugg: The example a few lines after demonstrates the shape of the object.
Could something like this be both understandable and picked up during docs parsing?
```
* @param {Object<eventName, {score: number, type: "conversion"|"click"|"view"}>} data.eventsScoring Scoring to apply to events
 ```

File: src/AlgoliaSearch.js (hunk 1 above)
Comment (julienbourdeau) [reply src/AlgoliaSearch.js]: I confirm this should be `setPersonalizationStrategy` and `getPersonalizationStrategy`. Thank you for bringing this up 🙏

File: src/AlgoliaSearch.js (hunk 3 above)
Comment (Haroenv) [reply src/AlgoliaSearch.js]: you need to go deeper like you did with data.eventsScoring, e.g. data.eventsScoring.eventName

File: src/AlgoliaSearch.js [hunk 4]
```diff
@@ -653,6 +653,57 @@ AlgoliaSearch.prototype.searchUserIDs = function(data, callback) {
...
+ * @return {Promise|undefined} Returns a promise if no callback given
+ * @example
+ * client.getPersonalizationStrategy();
+ */
+
+AlgoliaSearch.prototype.getPersonalizationStrategy = function(callback) {
+  return this._jsonRequest({
+    method: 'GET',
+    url: '/1/recommendation/personalization/strategy',
+    hostType: 'write',
```
julienbourdeau: Should be `read` I think

test/spec/common/client/test-cases/getPersonalizationStrategy.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = [
+  {
+    testName: 'client.getPersonalizationStrategy()',
+    object: 'client',
+    methodName: 'getPersonalizationStrategy',
+    callArguments: [],
+    action: 'write',
```
julienbourdeau: Here as well

src/AlgoliaSearch.js
Haroenv: not according to the REST API doc, that one mentioned algolia.net, which is the write endpoint :)

Haroenv: changed it to read in d7815c5d46db889032eb426cbb509e6ee89fa30a since I clarified with @marchelbling that this endpoint also works

PR #744, opened 2019-02-26
ci(test): use pinned bulkify version

Haroenv:
https://github.com/substack/bulkify/issues/19


Haroenv:
This changes the error message, but still errors unfortunately; cc @vvo

PR #745, opened 2019-02-26
test(ci): use version from nvm

vvo:


vvo:
@Haroenv I think this might be the actual issue we were having or a good part of it. Since Node.js 11 went out, our travis file said we should use it. Thus it failed because the whole dev stack is meant to be working with node.js 9.

PR #747, opened 2019-03-22
fix(ua): change the User-Agent to use the new specs lib (version)

robertmogos:
**Summary**
Update the UA to lib (${version})

### DIFF
--- a/src/browser/builds/algoliasearchLite.js
+++ b/src/browser/builds/algoliasearchLite.js
@@ -3,4 +3,4 @@
 var AlgoliaSearchCore = require('../../AlgoliaSearchCore.js');
 var createAlgoliasearch = require('../createAlgoliasearch.js');
 
-module.exports = createAlgoliasearch(AlgoliaSearchCore, '(lite) ');
+module.exports = createAlgoliasearch(AlgoliaSearchCore, 'lite ');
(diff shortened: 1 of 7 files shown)


src/browser/builds/algoliasearchLite.js
```diff
@@ -3,4 +3,4 @@
 var AlgoliaSearchCore = require('../../AlgoliaSearchCore.js');
 var createAlgoliasearch = require('../createAlgoliasearch.js');
 
-module.exports = createAlgoliasearch(AlgoliaSearchCore, '(lite) ');
+module.exports = createAlgoliasearch(AlgoliaSearchCore, 'lite ');
```
Haroenv: The trailing space was already present, but is it necessary? The jQuery one e.g. doesn't seem to have it

Haroenv: Ok I understand, it's to avoid the double space

src/server/builds/node.js
```diff
@@ -69,7 +69,7 @@ function algoliasearch(applicationID, apiKey, opts) {
 }
 
 algoliasearch.version = require('../../version.js');
-algoliasearch.ua = 'Algolia for Node.js ' + algoliasearch.version;
+algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + ')';
```
samouss: For this one we should add the Node version.

PR #749, opened 2019-03-27
fix(ua): Add the Node.js version in the UA

robertmogos:
Update the user agents to match the spec:

- Node: `Algolia for JavaScript (3.32.1); Node.js (8.15.1)`
- Browser: `Algolia for JavaScript (3.32.1); Browser`
- Browser lite: `Algolia for JavaScript (3.32.1); Browser (lite)`
- AngularJS: `Algolia for JavaScript (3.32.1); AngularJS (1.6.0)`
- jQuery: `Algolia for JavaScript (3.32.1); jQuery (3.2.1)`
- React Native: `Algolia for JavaScript (3.32.1); React Native`

### DIFF
--- a/src/browser/builds/algoliasearch.angular.js
+++ b/src/browser/builds/algoliasearch.angular.js
@@ -34,7 +34,11 @@ window.angular.module('algoliasearch', [])
     }
 
     algoliasearch.version = require('../../version.js');
-    algoliasearch.ua = 'Algolia for AngularJS (' + algoliasearch.version + ')';
+
+    algoliasearch.ua =
+      'Algolia for JavaScript (' + algoliasearch.version + '); ' +
+      'AngularJS (' + window.angular.version.full + ')';
+
     algoliasearch.initPlaces = places(algoliasearch);
 
     // we expose into window no matter how we are used, this will allow
--- a/src/browser/builds/algoliasearchLite.js
+++ b/src/browser/builds/algoliasearchLite.js
@@ -3,4 +3,4 @@
 var AlgoliaSearchCore = require('../../AlgoliaSearchCore.js');
 var createAlgoliasearch = require('../createAlgoliasearch.js');
 
-module.exports = createAlgoliasearch(AlgoliaSearchCore, 'lite ');
+module.exports = createAlgoliasearch(AlgoliaSearchCore, 'Browser (lite)');
--- a/src/server/builds/node.js
+++ b/src/server/builds/node.js
@@ -69,8 +69,11 @@ function algoliasearch(applicationID, apiKey, opts) {
 }
 
 algoliasearch.version = require('../../version.js');
-algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + '); ' +
-                  'Node.js (' + process.versions.node + ')';
+
+algoliasearch.ua =
+  'Algolia for JavaScript (' + algoliasearch.version + '); ' +
+  'Node.js (' + process.versions.node + ')';
+
 algoliasearch.initPlaces = places(algoliasearch);
 
 function AlgoliaSearchNodeJS(applicationID, apiKey, opts) {
--- a/src/server/builds/node.js
+++ b/src/server/builds/node.js
@@ -69,7 +69,8 @@ function algoliasearch(applicationID, apiKey, opts) {
 }
 
 algoliasearch.version = require('../../version.js');
-algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + ')';
+algoliasearch.ua = 'Node.js (' + process.versions["node"] + '); ';
+algoliasearch.ua += 'Algolia for Node.js (' + algoliasearch.version + ')';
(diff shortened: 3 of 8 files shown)


src/server/builds/node.js
```diff
@@ -69,7 +69,8 @@ function algoliasearch(applicationID, apiKey, opts) {
 }
 
 algoliasearch.version = require('../../version.js');
-algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + ')';
+algoliasearch.ua = 'Node.js (' + process.versions.node + '); ';
+algoliasearch.ua += 'Algolia for Node.js (' + algoliasearch.version + ')';
```
Haroenv:
```suggestion
algoliasearch.ua = 'Node.js (' + process.versions.node + '); Algolia for Node.js (' + algoliasearch.version + ')';
```

src/server/builds/node.js
```diff
@@ -69,7 +69,8 @@ function algoliasearch(applicationID, apiKey, opts) {
 }
 
 algoliasearch.version = require('../../version.js');
-algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + ')';
+algoliasearch.ua = 'Node.js (' + process.versions.node + '); ';
```
Haroenv:
```suggestion
```

robertmogos: I don't understand? Are you proposing to delete it ?

src/server/builds/node.js
robertmogos: I prefer mutating it. It is easier to understand that we actually have two user-agents.

samouss: We rarely use this kind of syntax `+=`. We avoid the mutation as much as possible. We can chain them in that case it won't be complexify that much. We can also invert the order makes more sense to have the `Algolia for ...` first.

src/server/builds/node.js
```diff
@@ -69,7 +69,8 @@ function algoliasearch(applicationID, apiKey, opts) {
 }
 
 algoliasearch.version = require('../../version.js');
-algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + ')';
+algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + ')' +
```
francoischalifour: `; ` should be on the first one.

src/server/builds/node.js
```diff
@@ -69,7 +69,8 @@ function algoliasearch(applicationID, apiKey, opts) {
 }
 
 algoliasearch.version = require('../../version.js');
-algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + ')';
+algoliasearch.ua = 'Algolia for Node.js (' + algoliasearch.version + '); ' +
```
francoischalifour:
```suggestion
algoliasearch.ua = [
  'Algolia for Node.js (' + algoliasearch.version + ')',
  'Node.js (' + process.versions.node + ')'
].join('; ');
```

This would actually be more readable and less error-prone.

samouss:
I've updated all the UA to be more compliant to the spec. I've updated the description of the PR with the output for each of the different build.

src/browser/builds/algoliasearch.angular.js
```diff
@@ -34,7 +34,11 @@ window.angular.module('algoliasearch', [])
     }
 
     algoliasearch.version = require('../../version.js');
-    algoliasearch.ua = 'Algolia for AngularJS (' + algoliasearch.version + ')';
+
+    algoliasearch.ua =
+      'Algolia for JavaScript (' + algoliasearch.version + '); ' +
+      'AngularJS (' + window.angular.version.full + ')';
```
Haroenv: can nothing of this be undefined?

samouss: `window.angular` is defined for sure otherwise the script has already thrown [a couple of lines above](https://github.com/algolia/algoliasearch-client-javascript/blob/5de2ccd84320ea967eb02b543d8501fdbb7a3a64/src/browser/builds/algoliasearch.angular.js#L24). The `version` is available on the `angular` object since [v0.9.18](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#0918-jiggling-armfat-2011-07-29).

src/browser/builds/algoliasearchLite.js
```diff
@@ -3,4 +3,4 @@
 var AlgoliaSearchCore = require('../../AlgoliaSearchCore.js');
 var createAlgoliasearch = require('../createAlgoliasearch.js');
 
-module.exports = createAlgoliasearch(AlgoliaSearchCore, 'lite ');
+module.exports = createAlgoliasearch(AlgoliaSearchCore, 'Browser (lite)');
```
Haroenv: I think we should avoid the parens here to have an easier to parse, but will leave the call up to @mthuret

samouss: Marie agreed on the format, we can move forward with it.

PR #766, opened 2019-07-01
chore(lint): add linting

sarahdayan:
**Summary**

This PR adds formatting and linting to the project:

- Prettier and ESLint (with the [Algolia configuration](https://github.com/algolia/eslint-config-algolia))
- EditorConfig

Linting can be performed manually. Additionally, every attempt to commit triggers linting (with Husky and lint-staged).

The idea is to ensure we don't commit anything "dirty" and avoid breaking a build because of linting errors. This reduces (if not completely removes) the need for "style" commits.

**Result**

To test this, try to:

- Add a JavaScript or TypeScript file with incorrect formatting or rule violations.
- Run `yarn lint`, this should raise errors.
- Stage the faulty file and try to commit it, this should raise errors as well.
- Run `yarn lint:fix`, this should take care of fixable issues.

lint

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -98,6 +98,13 @@
   dependencies:
     "@babel/helper-plugin-utils" "^7.0.0"
 
+"@babel/runtime@^7.0.0":
+  version "7.4.5"
+  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.4.5.tgz#582bb531f5f9dc67d2fcb682979894f75e253f12"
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -98,6 +98,13 @@
   dependencies:
```
nunomaduro: Weird, this file is on `.gitignore`.

Haroenv: yarn.lock should not be on gitignore, it provides a consistent dev environment for everyone

PR #778, opened 2019-09-04
feat: remaining-validity

### LINKED ISSUE #772: New method: GetSecuredApiKeyRemainingValidity
## Summary

When a secured API keys expire, search breaks. The current workaround is more complex
than it needs to be. 

The options for developers using Algolia require, at every place where a search
operation is performed, catching an operation error, identify it as a
"ValidUntil" error, then regenerate the API key, and perform the search
operation again.

The idea is to define a method which tells if a `SecuredAPIKey`
is expired or not without making a single call to Algolia's API.

## References

- Specs: https://github.com/algolia/algoliasearch-client-specs-internal/pull/25

nunomaduro:
This pull request adds `getSecuredApiKeyRemainingValidity` as specified on the api client specs.

closes #772

### DIFF
--- a/src/AlgoliaSearch.js
+++ b/src/AlgoliaSearch.js
@@ -732,6 +732,7 @@ AlgoliaSearch.prototype.disableRateLimitForward = notImplemented;
 AlgoliaSearch.prototype.useSecuredAPIKey = notImplemented;
 AlgoliaSearch.prototype.disableSecuredAPIKey = notImplemented;
 AlgoliaSearch.prototype.generateSecuredApiKey = notImplemented;
+AlgoliaSearch.prototype.getSecuredApiKeyRemainingValidity = notImplemented;
 
 function notImplemented() {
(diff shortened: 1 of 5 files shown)


Haroenv:
Looks good, bundle size change is odd though, since angular shouldn’t have this file, maybe it’s just the error

nunomaduro:
Waiting for https://github.com/algolia/algoliasearch-client-javascript/pull/779.

julienbourdeau:
It seems like this is only implemented for Node, not in the browser, am I correct?

If so, why is it not available in the browser?

nunomaduro:
Because the `generateSecuredApiKey` as also browser only. On the JavaScript on the default preset will contain this features on the browser too.

julienbourdeau:
I understand 👍 I'm very exited for v4.
This is a very important feature, as your frontend might be using secured api key, you need to know when to ask for a new key.

PR #781, opened 2019-09-17
chore(transporter): fix typo

Haroenv:
This isn't per se a typo, but `retryable` is "more common" https://english.stackexchange.com/questions/364277/which-spelling-is-preferred-retriable-or-retryable


Haroenv:
your bundlesize is set _very_ strict haha

nunomaduro:
Thanks!

Issue #755, opened 2019-05-15
[3.x] Method deleteObject deletes the index if objectID is empty


nunomaduro:
**What is the current behavior?**

Using `index.deleteObject` passing a empty string on the `objectID` deletes the index:

```js
var objectID = "";
index.deleteObject(objectID); // index is deleted.
```

**What is the expected behavior?**

Calling `index.deleteObject("");` should throw `AlgoliaSearchError` if `objectID` is a empty string.

nunomaduro:
Closing this, reasons described in #795.

PR #789, opened 2019-10-05
Fix/delete object with empty string

alexhoma:
**Summary**
Using `index.deleteObject` passing an empty string on the `objectID` deletes the index. And should throw an `AlgoliaSearchError`. Here's the linked issue: https://github.com/algolia/algoliasearch-client-javascript/issues/755

**Result**
Once repository was cloned, I've executed `yarn` to install dependencies and `yarn test`.

I've added a test to cover the "empty string" case and then I've added the guard clause on the production code. (https://github.com/algolia/algoliasearch-client-javascript/pull/789/commits/eadd2e35b74f3b6f1f593862317b0129c6387e8e)
Then, I've tried to refactor the checks of the `objectID`. I think I've improved legibility, but if not, we can avoid the last commit :) (https://github.com/algolia/algoliasearch-client-javascript/pull/789/commits/fb91860febba366f8d142a288a3cb58d1d5903ee)

**Notes**
- I'd like to discuss if we can add a test case when someone passes an empty string but its length is not 0, like: `"     "`. With the current code, this could not happen, but there is no test that prevents someone to break this.
- And another test to check if someone passes a number. Now I've seen that there is a guard that covers if `objectID` is not a number, but when I remove this condition, tests are green anyway.


alexhoma:
@nunomaduro  hi! integration tests are failing due to:
```
Error: missing: INTEGRATION_TEST_APPID=$APPID INTEGRATION_TEST_API_KEY=$APIKEY command.
```
Where should I add these params to work with Travis?

Haroenv:
@alexhoma, that's expected actually, the tests can not run for non-maintainers of the repo to avoid the API keys leaking

alexhoma:
Makes sense :sweat_smile: Thank you @Haroenv

nunomaduro:
@alexhoma Thanks for your contribution - had to revert this merge. The reason is because its introduces a behaviour breaking change when people perform `deleteObject(0);` -  this will throw an exception with the code introduced in this pull request.

Maybe you can re-perform this pull request with code that also covers this situation?

alexhoma:
@nunomaduro Sure I'll take it :ok_hand:

PR #790, opened 2019-10-08
getSecuredApiKeyRemainingValidity  has hanging comma, that throws error at runtime for node.js engine(s) <= 8

mattjacobs:
**Summary**
There's currently an issue for older node.js engines that do not support extra commas at the end of a method. 

**Result**
This method will now work for node.js clients


nunomaduro:
Thanks!

nunomaduro:
This version `3.35.1` contains this fix.

PR #795, opened 2019-10-16
[3.x] Fixes delete objects with empty string

alexhoma:
**Summary**
Reopened PR that prevents empty strings to be passed as an `objectId` when using `index.deleteObject`. The last PR was introducing a breaking change throwing an error when someone gaves a `0` as an objectId (and this was not valid). As explained in this comment: https://github.com/algolia/algoliasearch-client-javascript/pull/789#issuecomment-542104709

Related issue: https://github.com/algolia/algoliasearch-client-javascript/issues/755

**Result**
* Breaking change fixed. Now `0` is a valid object id.
* Also I've took the opportunity to test some uncovered cases: (https://github.com/algolia/algoliasearch-client-javascript/commit/c740877137240cc688b6fec36db557b338c0fd2d#diff-7f406cf06aa9340c5d564bde813b50f2)
  * When `deleteObject('id')` with a valid object id
  * When `deleteObject('id', cb)` with a valid object id and a callback


nunomaduro:
Updated the base of this branch.

nunomaduro:
@Haroenv Now that we have the v4 out, I would prefer to freeze the v3 completely. I was thinking about not merging this pull request as there is no actually bug on the method.

Haroenv:
I'm not sure if this isn't a bug, since v3 _does_ have safe programming. I'm ok with closing it though (definitely thank @alexhoma for your time making this PR), since it's fixed in v4

nunomaduro:
Yes - thanks @alexhoma !

PR #796, opened 2019-10-21
fix: unknown parameter `paginate` within findObjects method

nunomaduro:
**Summary**

This pull request fixes an issue when people try to find and object without pagination, setting `paginate` to false,  using the `findObject` method. Without the `delete` code introduced in this pull request, the engine returns: `Unknown parameter: paginate`.

### DIFF
--- a/src/Index.js
+++ b/src/Index.js
@@ -928,6 +928,9 @@ Index.prototype.findObject = function(findCallback, requestOptions, callback) {
   var paginate = requestOptions.paginate !== undefined ? requestOptions.paginate : true;
   var query = requestOptions.query !== undefined ? requestOptions.query : '';
 
+  delete requestOptions.paginate;
+  delete requestOptions.query;
+
   var that = this;
   var page = 0;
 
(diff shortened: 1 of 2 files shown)


nunomaduro:
Thanks for forcing test quality. Just added tests.

src/Index.js
```diff
@@ -928,6 +928,9 @@ Index.prototype.findObject = function(findCallback, requestOptions, callback) {
   var paginate = requestOptions.paginate !== undefined ? requestOptions.paginate : true;
   var query = requestOptions.query !== undefined ? requestOptions.query : '';
 
+  delete requestOptions.paginate;
+  delete requestOptions.query;
```
Haroenv: doesn't look like query is actually removed here according to the test, is that expected?

nunomaduro: Without the delete, the test will fail having this result:

```
# findObject: object not found with non matching query
not ok 3 should be equal
  ---
    operator: equal
    expected: |-
      '{"params":"query=Algolia&page=0"}'
    actual: |-
      '{"params":"query=Algolia&query=Algolia&page=0"}'
```

nunomaduro: This is expected, the query should passed within params, and not as a new key on the payload.

Haroenv: makes sense, I misread the variable names

PR #806, opened 2019-11-13
chore(prettier): use dedicated config

Haroenv:
**Summary**

This will make sure that if you use "format document" via vscode prettier extension works as expected

**Result**

arguments to prettier config are moved to their own file


Haroenv:
was causing other conflicts, let's leave as is

PR #951, opened 2020-01-15
feat(version): expose version on presets

Haroenv:
This is the API we have in v3, so some people might rely on this? It also makes sense to me to access it before instantiation. If you disagree, no hard feelings :)

### DIFF
--- a/packages/algoliasearch/src/__tests__/default.test.ts
+++ b/packages/algoliasearch/src/__tests__/default.test.ts
@@ -8,6 +8,11 @@ const algoliasearch = new TestSuite('search').algoliasearch;
 const client = algoliasearch('appId', 'apiKey');
 
 describe('default preset', () => {
+  it('exposes the version on the constructor', () => {
+    expect(algoliasearch.version).toBe(version);
+    expect(algoliasearch.version.startsWith('4.')).toBe(true);
+  });
+
   it('sets default headers', () => {
     expect(client.transporter.headers).toEqual({
       'content-type': 'application/x-www-form-urlencoded',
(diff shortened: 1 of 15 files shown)


francoischalifour:
I think we should avoid mutating our exports like this. I believe bundlers consider this as a side effect. Why not export `version` the normal way?

nunomaduro:
Context, with this pull request we allow to get the version in 3 different ways:
- 1 - module `import { version } from '@algolia/common'.
- 2 - instance `client.version`
- 3 - constructor: `algoliasearch.version`.

francoischalifour:
@nunomaduro Then it would make sense to also be able to import it like this:

```js
import { version } from 'algoliasearch';
// and
import { version } from 'algoliasearch/lite';
```

Wydt?

Haroenv:
>  I believe bundlers consider this as a side effect

I didn't think of that, do you have a link for that?

nunomaduro:
@francoischalifour @Haroenv Feedback: when you put `sideEffects: false` bundlers will assume that there is no side effects at all - He will not even search for side effects. 

Second, adding a property to an object (`algoliasearch`) exported by our module could never be a side effect.

nunomaduro:
@Haroenv I agree with @francoischalifour that we should just provide two ways of getting the version.

1. on the modular way `import { version } from '@algolia/common'.`.
2. on the preset way `import { version } from 'PRESET';`

If you agree, I can clean up this pull request, and adapt it.

francoischalifour:
@Haroenv I think it came up when talking with @samouss. I can't find many resources on this topic. Webpack's guide on [Tree Shaking](https://webpack.js.org/guides/tree-shaking/) is interesting but doesn't talk about what is considered as a side effect.

PR #996, opened 2020-02-05
chore: adds tests on url encoding

nunomaduro:
This pull request adds tests on the `encode` helper used to encode index name.


nunomaduro:
@aseure I will update this tests eventually if we have something on the CTS.

Issue #997, opened 2020-02-06
[4.x] Access to Local Storage can be blocked in Chrome


olafcm:
We updated to v4 of the client.
Since then we started to receive the following error for Chrome users:

> Failed to read the 'localStorage' property from 'Window': Access is denied for this document.

The issue happens here:

```
function u(e) {
  var t = e.localStorage || window.localStorage,
    r = "algoliasearch-client-js-".concat(e.key),
    n = function() {
      return JSON.parse(t.getItem(r) || "{}")
    };
```

More info: https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document

Is it possible for the client to detect if the LocalStorage is available? I.e. use: https://github.com/Modernizr/Modernizr/blob/master/feature-detects/storage/localstorage.js

Thank you

nunomaduro:
Thanks for reporting this issue, I will investigate this tomorrow morning. 👍

nunomaduro:
@olafcm Just to be clear, your search is working, but you got the `warning` on the chrome console right?

olafcm:
Yes, I tested it and search works. It probably only affects the search results caching.

nunomaduro:
Actually just don't preserve the state of the hosts after the refresh: https://github.com/algolia/algoliasearch-client-javascript/blob/d862a912d764655a415a9bffc25634d625ebe88d/packages/algoliasearch/src/builds/browserLite.ts#L49.

I will see if we can avoid the warning on the console. Otherwise everything seems to be working as expected.

Haroenv:
I wonder if adding a .catch which still throws, but without the default error message will avoid it showing on the console.

https://github.com/algolia/algoliasearch-client-javascript/blob/d862a912d764655a415a9bffc25634d625ebe88d/packages/cache-browser-local-storage/src/createBrowserLocalStorageCache.ts#L31

nunomaduro:
We are using `catch` on the `FallbackableCache`:  https://github.com/algolia/algoliasearch-client-javascript/blob/d862a912d764655a415a9bffc25634d625ebe88d/packages/cache-common/src/createFallbackableCache.ts#L21

That performs a `catch` on the driver to fallback to the next one. Leet me investigate this.

nunomaduro:
@olafcm can you also review the pull request that I just made?

nunomaduro:
[...]

PR #999, opened 2020-02-07
fix: dom-exception on access to localStorage

nunomaduro:
fix: #997 

The access to the localStorage was being performed on the constructor of the `localStorage` driver making the `algoliasearch` function throwing the following exception:

Screenshot 2020-02-07 at 11 37 50

This pull request delays the access to the localStorage to the async API provided by `get`, `set`, and `clear`, so the error can be `catch` - making the `fallbackable` driver proxy to the next driver in the chain.

https://github.com/algolia/algoliasearch-client-javascript/blob/d862a912d764655a415a9bffc25634d625ebe88d/packages/cache-common/src/createFallbackableCache.ts#L21


Haroenv:
Could you not do typeof `window.localStorage === object ? window.localStorage : {}`or something like this in the constructor, that way the error will still happen in the get right?

PR #1009, opened 2020-02-12
fix: missing cursor while using browseObjects

Martindelataille:
Api return ApiError when the request is the third without cursor.
Api Request has only 3 tries without cursor params

Message: 'you should pass cursor of current page to fetch next page'

PR #1011, opened 2020-02-13
fix: wrong behaviour on browsing to next page/cursor in browse methods

nunomaduro:
This pull request addresses found problems on `browse` related methods in some edge cases:

1. When an index is too big, the browsing using `pages` is no longer available by the API giving the following error: `you should pass cursor of the current page to fetch next page`. So we adapt the browsable promise saying that when the response as a cursor, we use the cursor for browsing to the next page. ` : https://github.com/algolia/algoliasearch-client-javascript/compare/fix/browse-methods-edge-cases?expand=1#diff-dc6449cb3a3afc56b777c2dd08bb26c9R34

2. On the page related browses ( synonyms and rules ), the `hitsPerPage` wasn't sent to Algolia. Causing Algolia to send always 20 synonyms per batch, and early stop on pagination as `shouldStop: response => response.hits.length < 1000,`. https://github.com/algolia/algoliasearch-client-javascript/compare/fix/browse-methods-edge-cases?expand=1#diff-2f2de2f87e60dfb956427b9e0bdc21a7R26

Let me know if you want me to explain those two cases orally.

closes #1009

### DIFF
--- a/packages/client-search/src/createBrowsablePromise.ts
+++ b/packages/client-search/src/createBrowsablePromise.ts
@@ -28,7 +28,7 @@ export function createBrowsablePromise<TObject>(
 
         /**
          * Finally, if the response contains a cursor, we browse to the next
-         * batch using that same cursor. Otherwise, we just use the tradicional
+         * batch using that same cursor. Otherwise, we just use the traditional
          * browsing using the page element.
          */
         if (response.cursor) {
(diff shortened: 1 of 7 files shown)


packages/client-search/src/createBrowsablePromise.ts
```diff
@@ -1,30 +1,45 @@
...
+         */
         if (options.shouldStop(response)) {
           return resolve();
         }
 
-        // eslint-disable-next-line functional/immutable-data
-        data.page++;
+        /**
+         * Finally, if the response contains a cursor, we browse to the next
+         * batch using that same cursor. Otherwise, we just use the tradicional
```
Haroenv:
```suggestion
         * batch using that same cursor. Otherwise, we just use the traditional
```

PR #1013, opened 2020-02-14
feat(rules-filter-promotes): adds field to rule type

nunomaduro:
Adds type `filterPromotes` to rule.


Haroenv:
can we add docs links to those types?

nunomaduro:
@Haroenv I don't find elegant have types with links inside of it. Let's discuss it later, and do a pull request adding links everywhere.

Haroenv:
this worked ok https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f2db20ef615bd88a605a9b8750ec8ba170e2415a/types/algoliasearch/index.d.ts#L1213

PR #1023, opened 2020-02-19
fix: header field x-algolia-user-id is not allowed by access-control-allow-headers

nunomaduro:
Multi-cluster methods using `x-algolia-user-id` header on `POST`|`DELETE` verbs may encounter the bug `header field x-algolia-user-id is not allowed by access-control-allow-headers` while using this methods on the browser.

This PR fixes that exact same issue passing `x-algolia-user-id` on the query parameter instead of the browser.

In addition,  we also increase the timeouts of the transporter layer to avoid flakiness, and we did reactivate the mcm test now that the mcm cluster is alive again.

### DIFF
--- a/packages/client-search/src/methods/client/removeUserID.ts
+++ b/packages/client-search/src/methods/client/removeUserID.ts
@@ -11,7 +11,7 @@ export const removeUserID = (base: SearchClient) => {
     const mappedRequestOptions = createMappedRequestOptions(requestOptions);
 
     // eslint-disable-next-line functional/immutable-data
-    mappedRequestOptions.headers['X-Algolia-User-ID'] = userID;
+    mappedRequestOptions.queryParameters['X-Algolia-User-ID'] = userID;
 
     return base.transporter.write(
       {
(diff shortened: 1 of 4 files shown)


packages/client-search/src/methods/client/removeUserID.ts
```diff
@@ -11,7 +11,7 @@ export const removeUserID = (base: SearchClient) => {
     const mappedRequestOptions = createMappedRequestOptions(requestOptions);
 
     // eslint-disable-next-line functional/immutable-data
-    mappedRequestOptions.headers['X-Algolia-User-ID'] = userID;
+    mappedRequestOptions.queryParameters['X-Algolia-User-ID'] = userID;
```
Haroenv: can we get into similar URL too long problems as with secured api key?

nunomaduro: Yes - but in reality that doesn't happen.

@aseure @chloelbn any thoughts on this remark?

PR #1024, opened 2020-02-19
feat: adds hasPendingMappings

### LINKED ISSUE #803: [4.x] Add has_pending_mapping
Please add the following method to `search_client`

```java
function has_pending_mappings(retrieve_mappings: boolean = false, opts: request_options) 
return has_pending_mappings_responses

struct has_pending_mappings_response {
	pending: boolean,
	clusters: map<string, string[]>
}
```

```java
{
  "method": "GET",
  "url": "/1/clusters/mapping/pending?getClusters={retrieve_mappings}",
  "hostType": "read"
}
```

nunomaduro:
This pull requests closes #803.

Requires #1023 needs to be merged first.


Haroenv:
can you change the base of the PR to be #1023 instead, so the diff is smaller? (stacked PR)

nunomaduro:
@Haroenv Good idea, done!

packages/client-search/src/types/HasPendingMappingsOptions.ts
```diff
@@ -0,0 +1,17 @@
+export type HasPendingMappingsOptions = {
+  /**
+   * If the clusters pending mapping state should be on the response.
+   *
+   * @alias getClusters
```
Haroenv: if this is an alias, why do we support both?

packages/client-search/src/methods/client/hasPendingMappings.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const hasPendingMappings = (base: SearchClient) => {
+  return (
+    requestOptions?: HasPendingMappingsOptions & RequestOptions
+  ): Readonly<Promise<HasPendingMappingsResponse>> => {
+    const { retrieveMappings, ...options } = requestOptions || {};
+
+    if (retrieveMappings === true) {
+      // eslint-disable-next-line functional/immutable-data
+      options.getClusters = true;
```
Haroenv: I don't see any unit tests for this, on purpose?

nunomaduro:
```
  await expect(client.hasPendingMappings({ retrieveMappings: true })).resolves.toMatchObject({
     clusters: expect.any(Object),
     pending: expect.any(Boolean),
   });
```

packages/client-search/src/types/HasPendingMappingsOptions.ts
nunomaduro: Actually, I was thinking about removing `alias` here, and marking `getClusters` as internal. WDYT?

packages/client-search/src/methods/client/hasPendingMappings.ts
Haroenv: that's an integration test, no?

packages/client-search/src/types/HasPendingMappingsOptions.ts
Haroenv: if they're just aliases, I'd even argue for not making getClusters an argument

packages/client-search/src/methods/client/hasPendingMappings.ts
nunomaduro: @Haroenv Indeed, but what is the point of having a unit test if the integration one already covers all the cases?

packages/client-search/src/types/HasPendingMappingsOptions.ts
nunomaduro: I will add getClusters  it as `@internal` so we have type here: https://github.com/algolia/algoliasearch-client-javascript/blob/858e43dfacef7d26d1a44b147741183b0e8677b3/packages/client-search/src/methods/client/hasPendingMappings.ts#L14

Issue #991, opened 2020-02-03
discussion: browse objects does not allow cursor-style paging


kopertop:
Previously, there was a method called `browse` that allowed us to start a browse request, get a response, and then call `browseFrom` to continue browsing when we'd processed that initial batch and were ready for another one. Now, it appears all we have an option for is `browseObjects`, which takes a function to call for each "batch".

This has several issues with our current workflow:

1. It is nothing like the "search" method
2. It does not allow "cancelling" a browse (such as if we've hit the limit of how many we want to process)
3. If batch operations take a while and are asynchronous, this could cause memory issues (imagine processing 100m records all at once without any delay between batches).

It **does** appear the `browseObjects` function allows passing in a `cursor`, but it has no option to return results and **not continue**.

Proposed solution:

Allow an option to `browseObjects` that lets us call it the "old" way, or provide a new function `browse` and `browseFrom` like the old library supported.

Haroenv:
This is not a permanent solution yet, but a way to get around this limitation right now, would be to use the underlying `createBrowsablePromise` directly, and modify the `shouldStop` condition:

https://github.com/algolia/algoliasearch-client-javascript/blob/e4c3f59afb3278b5466c7a5977b45268ef466479/packages/client-search/src/methods/index/browseObjects.ts#L19

Haroenv:
Could you clarify a bit more why you are browsing only to a certain number of results, and not the whole dataset, or using filters? This way we can see which changes would enable your use case best

kopertop:
Our specific use-case is for _limited_ exports. We allow some our customers to export up to 10k results (which is configurable per customer). By default we still want to keep the site-search to 1k, so we got around this by using `browse`.

It's also a very different callback pattern to have to include a callback function vs being able to respond in an async/await style format. We have a wrapper around algoliasearch that works for both the backend and frontend by applying some custom filters on top of any search we pass in, and searching multiple indexes and combining the results. It's nice to be able to have that work with both browse and search methods (which it was before, but isn't now).

nunomaduro:
@kopertop A solution like https://github.com/algolia/algoliasearch-client-javascript/pull/1029 would solve your problem?

djake:
I agree with @kopertop's concerns, specifically about the synchronous nature of the `batch` option signature.

[...]

PR #1029, opened 2020-02-20
feat(browse): make shouldStop public on the browsable methods

Haroenv:
Exposing `shouldStop` allows people to create a different reason to stop browsing, e.g. if they only need the first 10.000 results instead of the whole index.

see #991 for use case examples

### DIFF
--- a/packages/client-search/src/methods/index/browseObjects.ts
+++ b/packages/client-search/src/methods/index/browseObjects.ts
@@ -12,7 +12,7 @@ import {
 
 export const browseObjects = (base: SearchIndex) => {
   return <TObject>(
-    requestOptions?: SearchOptions & BrowseOptions<TObject> & RequestOptions 
+    requestOptions?: SearchOptions & BrowseOptions<TObject> & RequestOptions
   ): Readonly<Promise<void>> => {
     return createBrowsablePromise<TObject>({
       shouldStop: response => response.cursor === undefined,
--- a/packages/client-search/src/methods/index/browseObjects.ts
+++ b/packages/client-search/src/methods/index/browseObjects.ts
@@ -12,11 +12,11 @@ import {
 
 export const browseObjects = (base: SearchIndex) => {
   return <TObject>(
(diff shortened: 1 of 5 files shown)


packages/client-search/src/types/BrowseOptions.ts
```diff
@@ -1,8 +1,10 @@
+import { BrowseResponse } from '.';
 import { ObjectWithObjectID } from './ObjectWithObjectID';
 
 export type BrowseOptions<TObject> = {
   /**
    * The callback called for each batch of objects.
    */
   readonly batch?: (batch: ReadonlyArray<TObject & ObjectWithObjectID>) => any;
+  readonly shouldStop?: (response: BrowseResponse<TObject>) => boolean;
```
Haroenv: TODO: add doc comment

nunomaduro:
Let's keep this on hold.

nunomaduro:
In progress, going to expose a `shouldStop` option for this.

packages/client-search/src/methods/index/browseObjects.ts
```diff
@@ -12,11 +12,11 @@ import {
 
 export const browseObjects = (base: SearchIndex) => {
   return <TObject>(
-    requestOptions?: SearchOptions & BrowseOptions<TObject> & RequestOptions
+    requestOptions?: SearchOptions & BrowseOptions<TObject> & RequestOptions
```
Haroenv, line 15:
```suggestion
    requestOptions?: SearchOptions & BrowseOptions<TObject> & RequestOptions
```

packages/client-search/src/types/BrowseOptions.ts
Haroenv:
```suggestion
  /**
   * condition to stop browsing. By default this checks whether there's any more content to get
   */
  readonly shouldStop?: (response: BrowseResponse<TObject>) => boolean;
```

Issue #798, opened 2019-10-24
feat: add explain decompounding at query time


Ant-hem:
Please add the following modifications to both `query` and `searchReponse` objects

**Addition to existing JSON schemas**

Query: new optional parameter `explain` (or not new if Query Rules v2 was already implemented)

```java
{
  "explain": string[]?
}
```

SearchResponse: new `explain` response field (or not new if Query Rules v2 was already implemented)

```java
        "explain": {
            "match": {
                "alternatives": [
                    {
                        "type": string,
                        "words": string[],
                        "typos": int,
                        "offset": int,
                        "length": int
                    }
                 ]
            }
         }
```

Haroenv:
also just for definitely typed

PR #1036, opened 2020-02-25
feat: adds explain in search response

nunomaduro:
This pull request solves #798.

### DIFF
--- a/packages/client-search/src/types/SearchResponse.ts
+++ b/packages/client-search/src/types/SearchResponse.ts
@@ -38,4 +38,50 @@ export type SearchResponse<TObject = {}> = {
   readonly parsedQuery?: string;
   readonly userData?: any;
   readonly appliedRules?: ReadonlyArray<Readonly<Record<string, any>>>;
+
+  /**
+   * The explanation of the decompounding at query time.
+   */
+  readonly explain?: {
+    /**
+     * The explain query match.
+     */
+    readonly match: {
+      /**
+       * The explain query match alternatives.
+       */
+      readonly alternatives: ReadonlyArray<{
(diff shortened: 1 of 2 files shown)


packages/client-search/src/types/SearchResponse.ts
```diff
@@ -38,4 +38,50 @@ export type SearchResponse<TObject = {}> = {
...
   readonly appliedRules?: ReadonlyArray<Readonly<Record<string, any>>>;
+
+  /**
+   * The explanation of the decompounding at query time.
+   */
+  readonly explain?: {
+    /**
+     * The explain query match.
+     */
+    readonly match: {
```
nunomaduro: Explain is optional, but when exists, the it will **always** the following **REQUIRED** structure:

packages/client-search/src/types/SearchResponse.ts
```diff
@@ -38,4 +38,50 @@ export type SearchResponse<TObject = {}> = {
...
+         */
+        readonly length: number;
+      }>;
+    };
+
+    /**
+     * Query parameter reporting. Parameters are reported
+     * as a JSON object with one field per parameter.
+     */
+    readonly params?: Readonly<Record<string, any>>;
```
nunomaduro: I didn't managed to actually get this key back. I am missing something?

Haroenv: Isn’t the type here something close to QueryParams?

nunomaduro: The format is something like: `{ abTests: Object, final: Array, ... }`. It may change in the near future.

PR #1042, opened 2020-02-28
fix(browse): expose original error from browse request

Haroenv:
Wrapping the original function in a new Promise swallowed its errors, and thus if there was some kind of error, you'd get an "unhandled promise" error like this:

```
(node:52050) UnhandledPromiseRejectionWarning: #<Object>
(node:52050) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:52050) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

instead of something like:

```js
{
  name: 'ApiError',
  message: 'Index prod_dictionaries does not exist',
  status: 404,
  transporterStackTrace: [
    {
      request: [Object],
      response: [Object],
      host: [Object],
      triesLeft: 3
    }
  ]
}
```

<details>
<summary>code to get that error</summary>

```
const fs = require('fs')
const algoliasearch = require('algoliasearch')
const client = algoliasearch('xxx', 'xxx')
const index = client.initIndex('xxx_invalid_index')
let hits = []
index
  .browseObjects({
    batch: objects => (hits = hits.concat(objects)),
  })
  .then(() => {
    console.log('Finished!')
    console.log('We got %d hits', hits.length)
    fs.writeFile('browse.json', JSON.stringify(hits, null, 2), 'utf-8', err => {
      if (err) throw err
      console.log('Your index has been exported!')
    })
  })
  .catch(e => {
    console.log(e)
  })
```

</details>

### DIFF
--- a/packages/client-search/src/__tests__/integration/browsing.test.ts
+++ b/packages/client-search/src/__tests__/integration/browsing.test.ts
@@ -11,8 +11,17 @@ import { TestSuite } from '../../../../client-common/src/__tests__/TestSuite';
 
 const testSuite = new TestSuite('browsing');
 
-test(testSuite.testName, async () => {
-  const index = testSuite.makeIndex();
+let index: ReturnType<typeof testSuite.makeIndex>;
+let browse: (
+  method: (index: SearchIndex) => any,
+  options?: { hitsPerPage?: number }
+) => Promise<{
+  hitsCount: number;
+  batchCount: number;
+}>;
+
+beforeAll(async () => {
+  index = testSuite.makeIndex();
 
   // First lets create the index.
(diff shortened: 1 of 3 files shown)


Haroenv:
The errors on CI seem unrelated, right @nunomaduro ?

nunomaduro:
Tests on the CI fixed.

nunomaduro:
Good for me, debugging while CI is failing on browser testing ( never failed before ).

PR #1051, opened 2020-03-04
feat(search-response-docs): adds docs

nunomaduro:
This pull request adds docs to the search response type.

### DIFF
--- a/packages/client-search/src/types/SearchResponse.ts
+++ b/packages/client-search/src/types/SearchResponse.ts
@@ -1,42 +1,175 @@
 import { ObjectWithObjectID } from '.';
 
-// @todo add docs here.
 export type SearchResponse<TObject = {}> = {
+  /**
+   * The hits returned by the search.
+   *
+   * Hits are ordered according to the ranking or sorting of the index being queried.
+   */
   readonly hits: ReadonlyArray<TObject & ObjectWithObjectID>;
+
+  /**
+   * Index of the current page (zero-based).
+   */
   readonly page: number;
+
+  /**
+   * Number of hits returned (used only with offset)
+   */
   readonly length?: number;
+
+  /**
+   * The offset of the first hit to returned.
+   */
   readonly offset?: number;
+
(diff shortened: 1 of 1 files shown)


packages/client-search/src/types/SearchResponse.ts
```diff
@@ -1,42 +1,175 @@
...
+
+  /**
+   * Index of the current page (zero-based).
+   */
   readonly page: number;
+
+  /**
+   * Number of hits returned (used only with offset)
+   */
   readonly length?: number;
```
Haroenv: we could have a discriminated union here: `baseParams & (hitsPerPageParams | offsetParams)`

packages/client-search/src/types/SearchResponse.ts
```diff
@@ -1,42 +1,175 @@
...
+
+  /**
+   * Used to return warnings about the query.
+   */
   readonly message?: string;
+
+  /**
+   * The computed geo location.
+   *
+   * Format: ${lat},${lng}, where the latitude and longitude are expressed as decimal floating point number.
```
Haroenv:
```suggestion
   * Format: "lat,lng", where the latitude and longitude are expressed as decimal floating point number.
```

this is also clear enough imo

packages/client-search/src/types/SearchResponse.ts
nunomaduro: I don't understand your comment.

Haroenv: ` { hits, ...} & ({ page, hitsPerPage} | { offset, length })`

something like that

Haroenv:
Not necessary, but they show up in the editor when you're typing, so that's good for DX

packages/client-search/src/types/SearchResponse.ts
nunomaduro: I won't have any type like that yet. Let's leave it for maybe next time.

PR #1068, opened 2020-03-11
fix(types): remove readonly from responses

### LINKED ISSUE #1064: feat: remove `readonly` constraint from response related types
**TL; DR**: Go to every type that ends with `...Response.ts` and remove `readonly` keywords.

A good type-system prevents problems from happening - that's why, by default, the given params to a async function in the client most be *readonly*. So they can't be modified while the promise is being resolved in background. Here is an example:

```ts
client.search('foo', params).then(console.log);
params.foo = 'bar'; // TYPE-ERROR: because those params may still be used inside the `search` function.
```

In the other hand, having `readonly` responses prevents the user from doing manipulations to the response object, and that's is annoying:

```ts
const response await client.search('foo', params);
delete response.hits; // TYPE-ERROR: grrrrrr
const myCustomObject = response;
```

Since having `readonly` responses add no value to the user or to us, to fix this problem, go to every type that ends with `...Response.ts` and remove `readonly` keywords.

francoischalifour:
## Description

This closes #1064 by removing all `readonly` attributes from the response types.

## How

- Ignore the rule `functional/prefer-readonly-type` in all `*Response.ts` files
- Remove the `Readonly<>` and `ReadonlyArray<>` types

### DIFF
--- a/packages/client-recommendation/src/types/GetPersonalizationStrategyResponse.ts
+++ b/packages/client-recommendation/src/types/GetPersonalizationStrategyResponse.ts
@@ -2,22 +2,22 @@ export type GetPersonalizationStrategyResponse = {
   /**
    * Events scoring
    */
-  readonly eventsScoring: ReadonlyArray<{
-    readonly eventName: string;
-    readonly eventType: string;
-    readonly score: number;
+  eventsScoring: Array<{
+    eventName: string;
+    eventType: string;
+    score: number;
   }>;
 
   /**
    * Facets scoring
    */
-  readonly facetsScoring: ReadonlyArray<{
-    readonly facetName: string;
-    readonly score: number;
+  facetsScoring: Array<{
+    facetName: string;
+    score: number;
   }>;
 
   /**
    * Personalization impact
    */
-  readonly personalizationImpact: number;
+  personalizationImpact: number;
 };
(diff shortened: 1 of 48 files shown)


packages/client-recommendation/src/types/GetPersonalizationStrategyResponse.ts
```diff
@@ -2,22 +2,22 @@ export type GetPersonalizationStrategyResponse = {
   /**
    * Events scoring
    */
-  readonly eventsScoring: ReadonlyArray<{
-    readonly eventName: string;
-    readonly eventType: string;
-    readonly score: number;
+  eventsScoring: ReadonlyArray<{
```
Haroenv: this is still ReadOnlyArray, not Array

packages/client-search/src/types/BrowseResponse.ts
```diff
@@ -4,10 +4,10 @@ export type BrowseResponse<TObject> = {
   /**
    * The hits per page.
    */
-  readonly hits: ReadonlyArray<TObject & ObjectWithObjectID>;
+  hits: ReadonlyArray<TObject & ObjectWithObjectID>;
```
Haroenv: same here

packages/client-recommendation/src/types/GetPersonalizationStrategyResponse.ts
francoischalifour: Good catch – I checked with a regex now and it seems to be fine!

Issue #1078, opened 2020-03-13
If available, ensure that Analytics' regions are "us" and "de" (not "eu")


aseure:
We've recently discovered that at least one API client was exposing "eu" as a
valid region for the Analytics client implementation, which is wrong. Only the
following regions are available at the moment to the Analytics client:

 - "us"
 - "de"

Haroenv:
For now we allow any string there.

PR #1085, opened 2020-03-16
feat(analytics): narrow the type for region

Haroenv:
see https://github.com/algolia/algoliasearch-client-javascript/issues/1078

This PR isn't necessary, but still will be nice for writing it, since only those two strings will be suggested


nunomaduro:
Thanks for this!

PR #1086, opened 2020-03-16
fix(types): fix missing type exports

### LINKED ISSUE #1080: proposal: re-exports all types in `algoliasearch`
Today, types like `SearchOptions`, or `Rule` are not exported by the `algoliasearch` package. 

People are actually importing them directly from the modular packages:

```
import { SearchOptions } from '@algolia/client-search';
```

**Solution**: re-exports all types, and types from modular packages, in `algoliasearch` and configure API Extrator to warn us about non exported types.

yannickcr:
Fixes #1080

This change marks all `@algolia/*` packages as [`bundledPackages`](https://api-extractor.com/pages/commands/config_file/#bundledpackages) for `algoliasearch`, which means that they are treated as part of this package and not external dependencies.

As a result API Extractor will throw an error if all types from `@algolia/*` packages that are accessible publicly through `algoliasearch` are not properly re-exported.

This PR is separated into 2 commits:
* The 1st one updates API Extractor configuration to mark `ae-forgotten-export` messages as errors and updates the build script to mark all `@algolia/*` packages as `bundledPackages` (for `algoliasearch` only). This is done dynamically so we do not have to update the configuration if we add/remove some packages.
*  The 2nd one updates the actual TS files to fix the missing exports.


packages/algoliasearch/src/builds/node.ts
```diff
@@ -175,6 +175,142 @@ import { createUserAgent, RequestOptions } from '@algolia/transporter';
 
 import { AlgoliaSearchOptions, InitAnalyticsOptions, InitRecommendationOptions } from '../types';
 
+export {
```
nunomaduro: Note: Types are not being exported in `browser.ts`.

packages/algoliasearch/src/builds/browserLite.ts
```diff
@@ -26,6 +26,40 @@ import { createUserAgent, RequestOptions } from '@algolia/transporter';
 
 import { AlgoliaSearchOptions } from '../types';
 
+export {
```
nunomaduro: If eventually, a new modification introduces a new type that is not exported, api extractor will complain right?

packages/algoliasearch/src/builds/node.ts
yannickcr: Yes, from what I saw there is no declaration file created from  `browser.ts`.
The entrypoints for the API Extractor are only `node.d.ts` (`api-extractor.json`) and `browserLite.d.t` (`api-extractor-lite.json`)

packages/algoliasearch/src/builds/browserLite.ts
yannickcr: Exactly :slightly_smiling_face:

packages/algoliasearch/src/builds/node.ts
nunomaduro: I see, let it this way then.

nunomaduro:
Very last @yannickcr can you build a version locally, and try it with react IS? 👍

yannickcr:
Good call @nunomaduro, the new export of `Response` was messing with the algoliasearch v3 detection ( https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/types/algoliasearch.ts ). I renamed it to `RequesterResponse` (and `Request` to `RequesterRequest` to be coherent).

nunomaduro:
I will double check this pull request after my current task.

Haroenv:
need to do that change in helper & InstantSearch.js too then

yannickcr:
I renamed the exports it in the client ( https://github.com/algolia/algoliasearch-client-javascript/pull/1086/commits/c8710aec8b7b956ab1a523873f636c903d81e501#diff-e002f6d33e4974541b52b0359ef2f1f2R59-R64 ), so there is no need to do some changes in the Helper, React or IS.js

PR #1126, opened 2020-04-09
fix(apiKey): expose ApiKeyACL type

bodinsamuel:
- Add ApiKeyACL type so we can strictly type the ACL's of api keys

The list can be found here:
https://www.algolia.com/doc/guides/security/api-keys/#rights-acl


bodinsamuel:
@Ant-hem @aseure nuno asked me to ping you to see if you were interested to have this type in your client's language ☺️

packages/client-search/src/types/ApiKey.ts
```diff
@@ -0,0 +1,14 @@
+export type ApiKeyACL =
```
nunomaduro, line 1: You can you create a file named `ApiKeyAclType` with contents similar too: https://github.com/algolia/algoliasearch-client-javascript/blob/master/packages/client-search/src/types/BatchActionType.ts?

nunomaduro, line 1: You can you create a file named `ApiKeyAclType` with contents similar too: https://github.com/algolia/algoliasearch-client-javascript/blob/master/packages/client-search/src/types/BatchActionType.ts?

nunomaduro:
All tests passing.

Issue #1137, opened 2020-04-15
bug(logs): wrong type definitions


kombucha:
### Description

It seems like the type definitions for what a `Log` item is are wrong as of `algoliasearch@4.1.0`.
All the fields are defined as **camel case** when they should actually be **snake case**.

https://github.com/algolia/algoliasearch-client-javascript/blob/af659f18e5be53b9c0589aba087cd92026e4829d/packages/client-search/src/types/Log.ts#L5

eg:
- `queryHeaders` should be `query_headers`
- `timeStamp` should be`timestamp`
etc

I can make a PR :)

### Related
[See documentation](https://www.algolia.com/doc/api-reference/api-methods/get-logs/)
[Missing inner_queries field](https://github.com/algolia/algoliasearch-client-javascript/issues/1108)

nunomaduro:
Just made a query to the API and here is the response. Going to work on the fix tomorrow morning.

```
      timestamp: '2020-04-15T16:10:00Z',
      method: 'POST',
      answer_code: '200',
      query_body: '\n{\n  "query": "foo"\n}\n',
      answer: '\n' +
        '{\n' +
        '  "hits": [\n' +
        '    {\n' +
        '      "test contract": "foo",\n' +
        '      "_tags": [\n' +
        '        "App\\\\Article::99"\n' +
        '      ],\n' +
        '      "objectID": "App\\\\Article::99",\n' +
        '      "_highlightResult": {\n' +
        '        "test contract": {\n' +
        '          "value": "<em>foo</em>",\n' +
        '          "matchLevel": "full",\n' +
        '          "fullyHighlighted": true,\n' +
        '          "matchedWords": [\n' +
        '            "foo"\n' +
        '          ]\n' +
        '        },\n' +
        '        "_tags": [\n' +
        '          {\n' +
        '            "value": "App\\\\Article::99",\n' +
        '            "matchLevel": "none",\n' +
        '            "matchedWords": [\n' +
        '            ]\n' +
        '          }\n' +
        '        ]\n' +
        '      }\n' +
        '    },\n' +
        '    {\n' +
        '      "test contract": "foo",\n' +
        '      "_tags": [\n' +
        '        "App\\\\Article::98"\n' +
        '      ],\n' +
        '      "objectID": "App\\\\Article::98",\n' +
        '      "_highlightResult": {\n' +
        '        "test contract": {\n' +
        '          "value": "<em>foo</em>",\n' +
[...]

PR #1139, opened 2020-04-16
fix(types): Log types

chloelbn:
Related issue: https://github.com/algolia/algoliasearch-client-javascript/issues/1137

This fixes the types of Log, as well as putting some field optionals

### DIFF
--- a/packages/client-search/src/__tests__/integration/get-logs.test.ts
+++ b/packages/client-search/src/__tests__/integration/get-logs.test.ts
@@ -1,3 +1,5 @@
+import { Log } from '../..';
+import { RequiredKeys } from '../../../../client-common/src/__tests__/helpers';
 import { TestSuite } from '../../../../client-common/src/__tests__/TestSuite';
 
 const testSuite = new TestSuite('get_logs');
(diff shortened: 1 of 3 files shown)


nunomaduro:
Still not quite there looking at the tests:
```
    Expected: ArrayContaining ["timestamp", "method", "answer_code", "query_body", "answer", "url", "ip", "sha1", "query_headers", "nb_api_calls", "processing_time_ms"]
    Received: ["timestamp", "method", "answer_code", "query_body", "answer", "url", "ip", "query_headers", "sha1", "processing_time_ms", "index", "query_params"]
```

nunomaduro:
The bundledize is failing, but we didn't touch on the source code. I will try to check this.

packages/client-search/src/__tests__/integration/get-logs.test.ts
```diff
@@ -15,5 +17,18 @@ test(testSuite.testName, async () => {
     type: 'all',
   });
 
-  await expect(getLogsResponse.logs).toHaveLength(2);
+  expect(getLogsResponse.logs).toHaveLength(2);
+  const keys: Array<RequiredKeys<Log>> = [
```
Haroenv, line 21: there's a TS type `Required` that does this I think:

```suggestion
  const keys: Array<Required<Log>> = [
```

Haroenv: ah wait I might have misunderstood what the type is supposed to do, you want only the required keys?

nunomaduro: Yes, the array can only contain required keys.

PR #1140, opened 2020-04-17
chore(deps-and-node-lts): locks node lts and revert bump deps

nunomaduro:
This pull request locks the usage of the client in node LTS, and reverts the previous bump of dependencies.


package.json
```diff
@@ -30,17 +30,17 @@
...
+    "@babel/preset-typescript": "7.8.3",
+    "@microsoft/api-extractor": "7.7.8",
+    "@rollup/plugin-alias": "3.0.1",
     "@rollup/plugin-json": "4.0.2",
-    "@rollup/plugin-node-resolve": "7.1.3",
-    "@types/jest": "25.2.1",
-    "@types/node": "13.11.1",
-    "@typescript-eslint/eslint-plugin": "2.28.0",
-    "@typescript-eslint/parser": "2.28.0",
+    "@rollup/plugin-node-resolve": "7.1.1",
```
Haroenv, line 39: I suspect either of the rollup plugins to have caused the regression. Can you try upgrading one at a time to see which is causing it?

Otherwise it could be babel too

nunomaduro: Honestly, it requires a little bit more time than I have right now. So I probably prefer do it later so I can make a release next Monday.

PR #1144, opened 2020-04-21
chore(release): small tweaks

Haroenv:
- add contributing guide (based on IS.js one)
- fix small error in api-extractor.json
- better link in CI comment in release script
- replace default with defaultValue: https://api-extractor.com/pages/tsdoc/tag_defaultvalue/
- put todo in separate comments so they aren't found by api-extractor
- avoid using > & {} in tsdoc comments, they can be confused with html apparently, and this looks nicer than the backslashes

### DIFF
--- a/packages/client-search/src/types/HasPendingMappingsOptions.ts
+++ b/packages/client-search/src/types/HasPendingMappingsOptions.ts
@@ -2,14 +2,14 @@ export type HasPendingMappingsOptions = {
   /**
    * If the clusters pending mapping state should be on the response.
    *
-   * @default false
+   * @defaultValue false
    */
   readonly retrieveMappings?: boolean;
 
   /**
    * If the clusters pending mapping state should be on the response.
    *
-   * @default false
+   * @defaultValue false
    *
    * @internal
    */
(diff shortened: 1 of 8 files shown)


packages/client-search/src/types/HasPendingMappingsOptions.ts
```diff
@@ -2,14 +2,14 @@ export type HasPendingMappingsOptions = {
   /**
    * If the clusters pending mapping state should be on the response.
    *
-   * @default false
+   * @defaultValue false
```
nunomaduro, line 5: What is wrong with `@default`? I thought that was a valid ts-doc annotation.

nunomaduro:
Also seems that lint tests are failing.

packages/client-search/src/types/HasPendingMappingsOptions.ts
Haroenv: as far as I saw, @default is valid only in JSDoc, not in TSDoc + API Extractor as far as I could see: https://github.com/microsoft/tsdoc/issues/27

nunomaduro: https://api-extractor.com/pages/tsdoc/tag_defaultvalue/

nunomaduro: Thanks! ( I am sure you did ) make sure you update all `@default` to that new annotations.

Haroenv: yes, those are the only two

PR #1145, opened 2020-04-21
feat(cache): allow export & import

Haroenv:
In SSR, you usually want to do the same query on the server as on the client, however the query on the client is redundant, since the results are serialized.

The JS client needs elaborate workarounds right now (including aliasing the "search" method, like in [react-instantsearch](https://github.com/algolia/react-instantsearch/blob/c9666c0458aacf1bf0151f479f57de21032d68b7/packages/react-instantsearch-core/src/core/createInstantSearchManager.js#L341-L397)) to be able to initialize the memory cache with already done queries.

Another use case is for "prefilling" the cache for e.g. empty or really popular queries.

### DIFF
--- a/packages/cache-browser-local-storage/src/createBrowserLocalStorageCache.ts
+++ b/packages/cache-browser-local-storage/src/createBrowserLocalStorageCache.ts
@@ -26,10 +26,13 @@ export function createBrowserLocalStorageCache(options: BrowserLocalStorageOptio
       events: CacheEvents<TValue> = {
         miss: () => Promise.resolve(),
       }
-    ): Readonly<Promise<TValue>> {
+    ) {
       return Promise.resolve()
         .then(() => {
           const keyAsString = JSON.stringify(key);
+          // since TValue is needed for getNamespace to have the correct type
+          // we need to retype the whole get function.
+          // This could be fixed by making cache generic instead of its functions.
           const value = getNamespace<TValue>()[keyAsString];
 
           return Promise.all([value || defaultValue(), value !== undefined]);
@@ -40,7 +43,7 @@ export function createBrowserLocalStorageCache(options: BrowserLocalStorageOptio
         .then(([value]) => value);
     },
 
-    set<TValue>(key: object | string, value: TValue): Readonly<Promise<TValue>> {
+    set(key, value) {
       return Promise.resolve().then(() => {
         const namespace = getNamespace();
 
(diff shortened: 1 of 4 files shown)


packages/cache-common/src/types/Cache.ts
```diff
@@ -24,4 +24,9 @@ export type Cache = {
    * Clears the cache.
    */
   readonly clear: () => Readonly<Promise<void>>;
+
+  /**
+   * Retrieve the entries currently in cache.
+   */
+  readonly entries: () => Readonly<Promise<unknown>>;
```
Haroenv, line 31: could also be

```suggestion
  readonly entries: <TValue>() => Readonly<Promise<Record<string, TValue>>>;
```

packages/cache-browser-local-storage/src/createBrowserLocalStorageCache.ts
```diff
@@ -26,10 +26,13 @@ export function createBrowserLocalStorageCache(options: BrowserLocalStorageOptio
       events: CacheEvents<TValue> = {
         miss: () => Promise.resolve(),
```
Haroenv, line 27: these parameters need to stay because we are using the generic in getNameSpace. See https://twitter.com/haroenv/status/1252610543162179585

packages/cache-browser-local-storage/src/createBrowserLocalStorageCache.ts
```diff
@@ -64,10 +67,14 @@ export function createBrowserLocalStorageCache(options: BrowserLocalStorageOptio
...
     },
 
-    clear(): Readonly<Promise<void>> {
+    clear() {
       return Promise.resolve().then(() => {
         getStorage().removeItem(namespaceKey);
       });
     },
+
+    entries() {
```
eunjae-lee, line 76: Who's calling `entries()` ?

Haroenv: this method isn't strictly necessary, but this makes the format of the cache opaque in case you want to transfer from localStorage to another place. In this case it's mainly there for consistency with the memory cache.

In theory the local storage cache could also accept a entries as initialCache, like memory cache does now.

If we don't have an entries value, we have to rely on the format not changing between versions

packages/cache-common/src/types/Cache.ts
yannickcr, line 31: Seems you forgot the `entries` implementation in `packages/cache-common/src/createFallbackableCache.ts` and `packages/cache-common/src/createNullCache.ts`

Haroenv:
For now this isn’t the right approach

PR #1148, opened 2020-04-30
chore(circleci): adds badge

chloelbn:
Adds the CircleCI badge to the README


Haroenv:
Do we need this when the tests are so flaky? A “failed” badge isn’t a very good signal imo

nunomaduro:
Because of analytics/recommendation I really don't think our test suite will pass all the time. Sorry. Delaying this for later.

PR #1149, opened 2020-05-04
fix(batch): add missing BatchActionType

bodinsamuel:
Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cdffe8978f1bb97cba1e8eddb30580dd08456a54/types/algoliasearch/index.d.ts#L747-L754

### DIFF
--- a/packages/client-search/src/types/BatchActionType.ts
+++ b/packages/client-search/src/types/BatchActionType.ts
@@ -4,6 +4,8 @@ export const BatchActionEnum: Readonly<Record<string, BatchActionType>> = {
   PartialUpdateObject: 'partialUpdateObject',
   PartialUpdateObjectNoCreate: 'partialUpdateObjectNoCreate',
   DeleteObject: 'deleteObject',
+  DeleteIndex: 'delete',
+  ClearIndex: 'clear',
 };
 
 export type BatchActionType =
@@ -11,4 +13,6 @@ export type BatchActionType =
   | 'updateObject'
   | 'partialUpdateObject'
   | 'partialUpdateObjectNoCreate'
-  | 'deleteObject';
+  | 'deleteObject'
+  | 'delete'
+  | 'clear';


Haroenv:
Seems to be inconsistent with the docs: https://www.algolia.com/doc/api-reference/api-methods/batch/#method-param-action

Do the docs also need to be updated?

packages/client-search/src/types/BatchActionType.ts
```diff
@@ -4,11 +4,15 @@ export const BatchActionEnum: Readonly<Record<string, BatchActionType>> = {
   PartialUpdateObject: 'partialUpdateObject',
   PartialUpdateObjectNoCreate: 'partialUpdateObjectNoCreate',
   DeleteObject: 'deleteObject',
+  DeleteIndex: 'delete',
```
nunomaduro, line 7: @Ant-hem @aseure can you quick check those in your clients?

bodinsamuel:
For the ref, we are currently using that in the crawler so it works. Probably need a doc update indeed.

Example:
```js
var algoliasearch = require("algoliasearch");

const client = algoliasearch('<app>', '<key>');
await client.multipleBatch([
    {
      indexName: 'test-multiple-delete',
      action: 'addObject',
      body: { objectID: 'sample-objectID' },
    },
    {
      indexName: 'test-multiple-delete',
      action: 'clear',
      body: {},
    },
    {
      indexName: 'test-multiple-delete',
      action: 'delete',
      body: {},
    },
  ]);
```

packages/client-search/src/types/BatchActionType.ts
aseure: What do we need to check? If batch actions do accept `delete`? Could you open issues with `polyglot` with the exact context?

nunomaduro: @aseure Correct. I will do it.

aseure: @nunomaduro I've just checked and it's available in Go, Scala, Java, C#.

bodinsamuel: can we merge this one as everybody seems to agree ? ☺️

bodinsamuel:
@Haroenv can review it during the absence of nuno ? ☺️

PR #1151, opened 2020-05-07
feat(settings): add enablePersonalization type

### LINKED ISSUE #1150: Accept `enablePersonalization` boolean as a valid setting parameter
All API clients currently should accept the `enablePersonalization` boolean
as a valid search parameter, which defaults to `false` when missing. However,
@rayrutjes recently reported that this parameter is also part of the settings
of an index.

The needed change is to expose the `enablePersonalization` boolean as a
valid setting parameter, which defaults to `false` when missing. It still
needs to be accepted as a search parameter as well.

Haroenv:
fixes #1150

### DIFF
--- a/packages/client-search/src/types/Settings.ts
+++ b/packages/client-search/src/types/Settings.ts
@@ -273,7 +273,7 @@ export type Settings = {
    * Overrides Algolia's default normalization.
    */
   readonly customNormalization?: Readonly<Record<string, Readonly<Record<string, string>>>>;
-  
+
   /**
    * Enable personalization for queries by default
    */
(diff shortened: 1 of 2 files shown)


packages/client-search/src/types/Settings.ts
```diff
@@ -273,6 +273,11 @@ export type Settings = {
    * Overrides Algolia's default normalization.
    */
   readonly customNormalization?: Readonly<Record<string, Readonly<Record<string, string>>>>;
+  
+  /**
+   * Enable personalization for queries by default
+   */
+  readonly enablePersonalization?: boolean;
```
nunomaduro, line 280: In addition, we need to update the CTS: https://github.com/algolia/algoliasearch-client-specs-internal/pull/117/files.

PR #1154, opened 2020-05-18
fix(account-client-copy): fixes unhandled promise issue

### LINKED ISSUE #1153: accountCopyIndex has unhandled promise rejection
Using v4.2.0 of the JS API Client

Here is my full code:
```javascript
const algoliasearch = require('algoliasearch')
const accountCopyIndex = require('@algolia/client-account').accountCopyIndex

const srcAppId = 'F6ZK2OJ8P7'
const dstAppId = 'RSBCBF0EG8'
const srcApiKey = 'ask_me_for_the_key' // browse + settings
const dstApiKey = 'ask_me_for_the_key' // addObject + editSettings + settings
const srcClient = algoliasearch(srcAppId, srcApiKey)
const dstClient = algoliasearch(dstAppId, dstApiKey)

const src = srcClient.initIndex('TPT_Resources')
const dst = dstClient.initIndex('TPT_Resources_copy')
console.log('Copying', src.indexName, '→', dst.indexName)
accountCopyIndex(src, dst)
  .then(() => console.log('Done'))
  .catch(e => console.log(e))
```

And here is what I get when I run this code:
```
Copying TPT_Resources → TPT_Resources_copy
(node:25405) UnhandledPromiseRejectionWarning: #<Object>
(Use `node --trace-warnings ...` to show where the warning was created)
(node:25405) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:25405) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Tried with different apps and indices of different sizes (several millions, or a few hundred thousand records), and the result is the same. It did work for very small indices like 1-2k records.

nunomaduro:
This pull request fixes the issue that promises rejections weren't being bubble up to the end-user when calling `accountCopyIndex(source, destination)` without an `wait` method. It also contains some changes related to testing so we can ensure the good behaviour of this pull request.

Fixes #1153 

1. Tests were added to test `accountCopyIndex` with and without `wait` calls at the end.
2. Also, added a test when `accountCopyIndex` fails in middle, to make sure it bubbles up the error correctly.
3. Finally, it stops comparing rules strictly, and the API Introduced the `_metadata` field on getRule.

### DIFF
--- a/packages/client-account/src/methods/accountCopyIndex.ts
+++ b/packages/client-account/src/methods/accountCopyIndex.ts
@@ -64,7 +64,7 @@ export const accountCopyIndex = (
      * we need to resolve that array of async responses using a
      * `Promise.all`, and then resolve `void` for the end-user.
      */
-    promise.then(() => Promise.all(responses)).then(() => Promise.resolve()),
+    promise.then(() => Promise.all(responses)).then(() => undefined),
 
     /**
      * Next, if the end-user calls the `wait` method, we need to also call
--- a/packages/client-account/src/methods/accountCopyIndex.ts
+++ b/packages/client-account/src/methods/accountCopyIndex.ts
@@ -58,7 +58,20 @@ export const accountCopyIndex = (
     )
     .then(() => Promise.resolve());
 
-  return createWaitablePromise(promise, () =>
-    Promise.all(responses.map(response => response.wait()))
+  return createWaitablePromise(
+    /**
+     * The original promise will return an array of async responses, now
+     * we need to resolve that array of async responses using a
+     * `Promise.all`, and then resolve `void` for the end-user.
+     */
+    promise.then(() => Promise.all(responses)).then(() => Promise.resolve()),
+
+    /**
+     * Next, if the end-user calls the `wait` method, we need to also call
+     * the `wait` method on each element of of async responses.
+     */
+    (_response, waitRequestOptions) => {
+      return Promise.all(responses.map(response => response.wait(waitRequestOptions)));
+    }
   );
 };
--- a/packages/client-account/src/methods/accountCopyIndex.ts
+++ b/packages/client-account/src/methods/accountCopyIndex.ts
@@ -55,8 +55,7 @@ export const accountCopyIndex = (
         // eslint-disable-next-line functional/immutable-data
         batch: objects => responses.push(saveObjects(destination)(objects, requestOptions)),
       })
-    )
-    .then(() => Promise.resolve());
+    );
 
   return createWaitablePromise(
     /**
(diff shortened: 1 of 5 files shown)


packages/client-account/src/methods/accountCopyIndex.ts
```diff
@@ -58,7 +58,20 @@ export const accountCopyIndex = (
     )
     .then(() => Promise.resolve());
 
-  return createWaitablePromise(promise, () =>
```
nunomaduro, line 61: The problem as on this line, that as resolving the original `promise` only fortting about the promises inside of the `responses` variable.

packages/client-account/src/methods/accountCopyIndex.ts
```diff
@@ -58,7 +58,20 @@ export const accountCopyIndex = (
...
-    Promise.all(responses.map(response => response.wait()))
+  return createWaitablePromise(
+    /**
+     * The original promise will return an array of async responses, now
+     * we need to resolve that array of async responses using a
+     * `Promise.all`, and then resolve `void` for the end-user.
+     */
+    promise.then(() => Promise.all(responses)).then(() => Promise.resolve()),
```
Haroenv, line 67:
```suggestion
    promise.then(() => Promise.all(responses)).then(() => undefined),
```

this is equivalent

nunomaduro: Done.

packages/client-account/src/methods/accountCopyIndex.ts
```diff
@@ -58,7 +58,20 @@ export const accountCopyIndex = (
     )
     .then(() => Promise.resolve());
```
Haroenv, line 59:
```suggestion
    .then(() => undefined);
```

Haroenv, line 59: or even 
```suggestion
    );
```

(because you're chaining it later)

packages/client-account/src/methods/accountCopyIndex.ts
```diff
@@ -58,7 +58,20 @@ export const accountCopyIndex = (
     )
```
Haroenv, line 58: the steps above here don't all need to happen in separate promise steps, but could be done in a Promise.all (unless you have a reason to want this to happen sequentially

nunomaduro: this to happen sequentially

nunomaduro: They need to happen sequentially.

Issue #1155, opened 2020-05-20
Expose rule.consequence.promote.objectIDs string array


aseure:
On top of the existing `objectID` (singular) string field, a new `objectIDs`
(plural) string array field is now be exposed in the `rule.consequence.promote`
Rule object representation. This lets the user easily promote a full group of
objects without letting some priority conflicts to shift some objects at the end
of the promoted object list.

Haroenv:
looks like we only accept `objectID` (singular) now: https://github.com/algolia/algoliasearch-client-javascript/blob/master/packages/client-search/src/types/Rule.ts#L84

nunomaduro:
This feature can't be merged yet: algolia/algoliasearch-client-specs-internal#118.

PR #1156, opened 2020-05-20
feat(rule): add group promotes

Haroenv:
doesn't have a test yet, but allows you to base on top of this.

fixes https://github.com/algolia/algoliasearch-client-javascript/issues/1155

see: https://www.algolia.com/doc/api-reference/api-methods/save-rule/#method-param-objectids

### DIFF
--- a/packages/client-search/src/types/Rule.ts
+++ b/packages/client-search/src/types/Rule.ts
@@ -93,7 +93,7 @@ export type ConsequencePromote =
       /**
        * List of unique identifiers for the objects to promote.
        */
-      readonly objectIDs: string[];
+      readonly objectIDs: readonly string[];
 
       /**
        * Promoted start rank for the objects (zero-based).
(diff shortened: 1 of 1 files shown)


packages/client-search/src/types/Rule.ts
```diff
@@ -77,17 +77,29 @@ export type ConsequenceQuery = {
...
+  | {
+      /**
+       * List of unique identifiers for the objects to promote.
+       */
+      readonly objectIDs: string[];
+
+      /**
+       * Promoted start rank for the objects (zero-based).
+       */
+      readonly position: number;
```
Haroenv, line 101: in principle we could have this in a shared type, but I wanted a different description for position if matched with an array

nunomaduro:
This feature can't be merged yet: https://github.com/algolia/algoliasearch-client-specs-internal/issues/118.

PR #1162, opened 2020-06-07
fix(requester-node-http): fix handling chunked responses (#1161)

### LINKED ISSUE #1161: bug: unicode characters in responses are sometimes corrupted v4 (node)
Response text that contains unicode characters is corrupted on large responses when using within node. This behaviour can reproduced in algoliasearch v4, version 3 works fine. 

I boiled the problem down to the `NodeHttpRequester` which simply adds buffer chunks to a string which leads to problems if the buffer only contains part of the unicode character. e.g.:

```
> a = Buffer.from("öäü")
<Buffer c3 b6 c3 a4 c3 bc>
> b = Buffer.from([0xc3, 0xb6, 0xc3])
<Buffer c3 b6 c3>
> c = Buffer.from([0xa4, 0xc3, 0xbc])
<Buffer a4 c3 bc>
> d = "" + b + c;  // <--- This is the current implementation in @algolia/node-http-requester
'ö��ü'
> Buffer.concat([b,c]);
<Buffer c3 b6 c3 a4 c3 bc>
> Buffer.concat([b,c]).toString()
'öäü'
```

I already created a fix and will make a pull request shortly.

vschoettke:
Fix #1161

### DIFF
--- a/packages/requester-node-http/src/createNodeHttpRequester.ts
+++ b/packages/requester-node-http/src/createNodeHttpRequester.ts
@@ -27,10 +27,12 @@ export function createNodeHttpRequester(): Requester & Destroyable {
         };
 
         const req = (url.protocol === 'https:' ? https : http).request(options, response => {
-          // eslint-disable-next-line functional/no-let
-          let content = '';
+          // eslint-disable-next-line functional/no-let, functional/prefer-readonly-type
+          let contentBuffers: Buffer[] = [];
 
-          response.on('data', chunk => (content += chunk));
+          response.on('data', chunk => {
+            contentBuffers = contentBuffers.concat(chunk);
+          });
 
           response.on('end', () => {
             // eslint-disable-next-line @typescript-eslint/no-use-before-define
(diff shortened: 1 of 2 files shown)


packages/requester-node-http/src/createNodeHttpRequester.ts
```diff
@@ -40,7 +42,7 @@ export function createNodeHttpRequester(): Requester & Destroyable {
 
             resolve({
               status: response.statusCode || 0,
-              content,
+              content: Buffer.concat(contentBuffers).toString(),
```
Haroenv, line 45: I checked, and this seems to be the same as we had in v3: https://github.com/algolia/algoliasearch-client-javascript/blob/v3.20.3/src/server/builds/node.js#L187

nunomaduro: Thanks for that info, just finishing up shouldStop issue, and I will jump into this.

PR #1165, opened 2020-06-09
chore(tests): tests against node 14

nunomaduro:
Node version 14 is available since the 21/04/2020: [medium.com/@nodejs/node-js-version-14-available-now-8170d384567e](https://medium.com/@nodejs/node-js-version-14-available-now-8170d384567e)

This pull request tests this library against that version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,6 +133,8 @@ workflows:
           version: '12'
       - test_unit:
           version: '13'
+      - test_unit:
+          version: '14'
       - test_lint:
           version: '12'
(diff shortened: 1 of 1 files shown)


yannickcr:
While adding "stable" would warn us if we are not compatible with the latest Node.js version it would also test against a moving target, making the tests more likely to fail without any apparent reason.

Would be nice if we could test against "stable" but also allow tests failure on it.

nunomaduro:
@yannickcr @Haroenv Can you perform a different pull request on that? Not sure I got it.

PR #1166, opened 2020-06-09
fix(log-type-inner-queries): apply snake_case

nunomaduro:
Hey @Ant-hem , can you confirm that the type is now correct? Including what is optional or not:

```ts
  inner_queries: Array<{
    index_name: string; // required
    query_id?: string; // optional
    offset?: number; // optional
    user_token?: string; // optional
  }>;
```

### DIFF
--- a/packages/client-search/src/types/Log.ts
+++ b/packages/client-search/src/types/Log.ts
@@ -91,6 +91,6 @@ export type Log = {
     /**
      * The user token of the query.
      */
-    readonly user_token: string;
+    readonly user_token?: string;
   }>;
 };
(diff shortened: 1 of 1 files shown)


packages/client-search/src/types/Log.ts
```diff
@@ -91,6 +91,6 @@ export type Log = {
     /**
      * The user token of the query.
      */
-    readonly userToken: string;
```
Ant-hem, line 94: userToken is optional

nunomaduro: You mean `user_token` right?

nunomaduro: Fixed, can you validate the pull request please?

Ant-hem: Yes, `user_token` sorry.

Issue #1170, opened 2020-06-18
Multi-Condition Rules feature


aseure:
This feature brings the support for multi-condition rules, which literally
means you can pass multiple conditions to trigger a single rule. The change
is backward-compatible and would ideally be shipped by mid-July or end-of-July
and it's gonna be deployed by June 19th for us to implement.

The feature is about exposing a new `conditions` array of rule conditions
in the rule object. The existing `condition` field remains for
backward-compatibility.

Also, please update the CTS like so: https://github.com/algolia/algoliasearch-client-specs-internal/pull/126/files.

PR #1171, opened 2020-06-25
feat: multiple conditions on query rules

nunomaduro:
This pull request adds the possibility of passing multiple conditions on query rules.

Closes https://github.com/algolia/algoliasearch-client-javascript/issues/1170.

### DIFF
--- a/packages/client-search/src/types/Rule.ts
+++ b/packages/client-search/src/types/Rule.ts
@@ -12,6 +12,11 @@ export type Rule = {
   readonly condition?: Condition;
 
   /**
+   * Conditions of the rule, expressed using the following variables: pattern, anchoring, context.
+   */
+  readonly conditions?: readonly Condition[];
+
+  /**
    * Consequence of the rule. At least one of the following object must be used: params, promote, hide, userData.
    */
   readonly consequence?: Consequence;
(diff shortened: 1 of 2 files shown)


packages/client-search/src/types/Rule.ts
```diff
@@ -11,6 +11,11 @@ export type Rule = {
    */
   readonly condition?: Condition;
 
+  /**
+   * Conditions of the rule, expressed using the following variables: pattern, anchoring, context.
+   */
+  readonly conditions?: readonly Condition[];
```
Haroenv, line 17: It should be a disjunct union: `{ condition } | { conditions }` with required in either case. That way you can’t have an object with both condition & conditions

nunomaduro: @Haroenv None of them are actually required.

However, we normally don't go that far with types. What do you think @Ant-hem @aseure?

Haroenv: What does a query rule without condition mean? One of them is required, no?

nunomaduro: @Haroenv No. You can the only required files are: `objectID`, and `consequence`. This is quite often used by people that want to apply a query rule all the time.

If you agree, I would to prefer to keep the rule typing just simple like this. We usually don't go that far with types, like possible combinations of key in rules, etc.

aseure: Agree with @nunomaduro on this: we don't go that far as enforcing the mutual exclusion here since the engine does not either.

PR #1179, opened 2020-07-23
test(perso): work around limitations on setPersonalizationStrategy

### LINKED ISSUE #1178: test: prevent error in CTS personalization test on 429 HTTP code
The personalization API is now limiting the number of
`set_personalization_strategy()` successful calls to 15 per day. Because of
that, our integration tests against the real API is not returning the following
error very often on our testing application:

```json
{
    "status": 429,
    "message": "Number of strategy saves exceeded for the day"
}
```

Because of that, if the CTS is implemented in the client, we need to perform
the `set_personalization_strategy()` call in our integration test like before
but to prevent the test from failing if the 429 error is returned (considered
it as a "success" and moving on with the `get_personalization_strategy()` call).

Haroenv:
fixes #1178

Basically the perso API has a limit of 15 calls per day, per application, which will make all tests fail. We prepare ourselves for this now, and also accept the 429 response to be correct

### DIFF
--- a/packages/client-recommendation/src/__tests__/features/personalization-strategy.test.ts
+++ b/packages/client-recommendation/src/__tests__/features/personalization-strategy.test.ts
@@ -51,7 +51,7 @@ test(testSuite.testName, async () => {
     expect(response).toEqual(successResponse);
   } catch (error) {
     // eslint-disable-next-line jest/no-try-expect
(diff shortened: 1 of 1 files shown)


Haroenv:
There's more context on slack (it's still an expensive operation, even if only for our test application), and we only accept this specific 429 (anything else will still error), so I think this makes the most sense

PR #1180, opened 2020-07-29
feat(createNodeHttpRequester): Add proxy support to another Agent on createNodeHttpRequester

diogolmenezes:
Hey guys, today i needed call Algolia behind my entherprise proxy server.

I did not found no way to do this with this lib so I added to the createNodeHttpRequester this feature.

To use it, you need to pass the proxy url to requester option.

https://www.algolia.com/doc/api-client/advanced/configure-the-client/javascript/?language=javascript#requester

like this:

const algoliasearch = algoliasearch(
  'YourApplicationID',
  'YourSearchOnlyAPIKey',
  {
    requester: createNodeHttpRequester('http://yourproxy:8080')
  },
);

### DIFF
--- a/packages/requester-node-http/package.json
+++ b/packages/requester-node-http/package.json
@@ -17,6 +17,7 @@
     "dist"
   ],
   "dependencies": {
-    "@algolia/requester-common": "4.3.1"
+    "@algolia/requester-common": "4.3.1",
+    "https-proxy-agent": "^5.0.0"
   }
 }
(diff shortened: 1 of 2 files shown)


Haroenv:
This makes sense, however this would make that package a a dependency, alternatively you could make the whole agent an argument, so you can choose the version, as well as choosing a different agent. What do you think?

diogolmenezes:
> This makes sense, however this would make that package a a dependency, alternatively you could make the whole agent an argument, so you can choose the version, as well as choosing a different agent. What do you think?

Nice, i will do it !

diogolmenezes:
> This makes sense, however this would make that package a a dependency, alternatively you could make the whole agent an argument, so you can choose the version, as well as choosing a different agent. What do you think?

Done. Now to use it, you have to :

`
const agent = new HttpsProxyAgent(process.env.http_proxy); 
const algoliasearch = algoliasearch(
          'YourApplicationID',
          'YourSearchOnlyAPIKey',
          {
                    requester: createNodeHttpRequester(agent)
          },
);`

PR #1183, opened 2020-08-18
fix(getObject): null in return type for items

dispix:
Using `getObjects` returns an array of records but also `null` when the object id doesn't match any record. Updating the type to reflect that.

See example of a response where the first object id matches a record and the second (object id "z") doesn't:
image


Haroenv:
seems like it's circle which is partially down, not the tests themselves per se. Can you do a force push?

PR #1184, opened 2020-08-19
feat(requestOptions): accept requesterOptions

Haroenv:
Sometimes (e.g. local engine, passing ca) it's needed to pass arguments to http[s].request. Since all keys are serialised into data / queryparameters, we need a new excluded key (`requesterOptions`) to pass arguments to `request`


Haroenv:
Too much hassle with the typings / bundlesize / tests. Will redo this if I face the issue again.

PR #1189, opened 2020-09-07
feat(hosts): accept url as string

### LINKED ISSUE #1188: We should support passing an array of string for the `hosts` options
I think we should either support or throw an exception when using an array of string for the `hosts` option:

```js
const searchClient = algoliasearch("", "", {
  hosts: ["test.com"]
});
```

instead of 

```js
const searchClient = algoliasearch("", "", {
  hosts: [{ url: "test.com/sitesearch" }]
});
```

Today it's creating a weird "undefined" domain.

Haroenv:
In v3 the API was `hosts: string[]`. It's easy to miss out on this in the migration guide, and it's also simpler to write just a string in the general case

fixes #1188

### DIFF
--- a/packages/transporter/src/createStatelessHost.ts
+++ b/packages/transporter/src/createStatelessHost.ts
@@ -1,6 +1,14 @@
 import { CallEnum, HostOptions, StatelessHost } from '.';
 
 export function createStatelessHost(options: HostOptions): StatelessHost {
+  if (typeof options === 'string') {
+    return {
+      protocol: 'https',
+      url: options,
+      accept: CallEnum.Any,
+    };
+  }
+
   return {
     protocol: options.protocol || 'https',
     url: options.url,
--- a/package.json
+++ b/package.json
@@ -97,11 +97,11 @@
   "bundlesize": [
     {
       "path": "packages/algoliasearch/dist/algoliasearch.umd.js",
-      "maxSize": "7.53KB"
+      "maxSize": "7.56KB"
     },
     {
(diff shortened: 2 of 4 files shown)


packages/transporter/src/createStatelessHost.ts
```diff
@@ -1,6 +1,14 @@
 import { CallEnum, HostOptions, StatelessHost } from '.';
 
 export function createStatelessHost(options: HostOptions): StatelessHost {
+  if (typeof options === 'string') {
+    return {
+      protocol: 'https',
```
redox, line 6: Do we need to force the protocol? I think I would rather keep not set the `protocol` key and let the regular business logic deal with it.

redox (commented):
Not sure about the `protocol`, what do you think?

packages/transporter/src/createStatelessHost.ts
Haroenv: this is the only place where protocol is set. If you set `{ url: xxx }` the protocol will also be https. If you want a custom protocol you can use the object form as normal. Don't you think that most cases would be using https?

yannickcr (approved):

packages/transporter/src/createStatelessHost.ts
yannickcr: I think `https` is a reasonable default here.
Also FMPOV the DX is better if `createStatelessHost({url: xxx})` and `createStatelessHost(xxx)` have the same behavior

PR #1190, opened 2020-09-07
chore(TypeScript): avoid ts-ignore

Haroenv:
- addMethods with `as`
- updateApiKey with type guard
- browseRules & browseSynonyms with `as any` (the key exists, but isn't supposed to be shown to the user, thus isn't in the type


packages/client-search/src/methods/client/updateApiKey.ts
```diff
@@ -35,17 +35,16 @@ export const updateApiKey = (base: SearchClient) => {
       'description',
       'maxQueriesPerIPPerHour',
       'maxHitsPerQuery',
-    ];
+    ] as const;
```
eunjae-lee, line 38: just curiosity, there's no mutation on `apiKeyFields` below, and why `as const`?

Haroenv: without `as const`, the type of `updatedField` lower is string, not the actual field values

eunjae-lee: Ah I didn't know. Thanks!

eunjae-lee (approved):
This looks good to me.

PR #1191, opened 2020-09-08
feat(types): introduce generic `Hit` type

francoischalifour:
This introduces a new generic `Hit` type that infers values for records' highlighting and snippeting results.

### DIFF
--- a/packages/client-search/src/types/Hit.ts
+++ b/packages/client-search/src/types/Hit.ts
@@ -5,7 +5,7 @@ type HighlightMatch = {
   readonly fullyHighlighted?: boolean;
 };
 
-type HighlightResult<THit> = THit extends string | number
+export type HighlightResult<THit> = THit extends string | number
   ? HighlightMatch
   : {
       [KAttribute in keyof THit]: HighlightResult<THit[KAttribute]>;
@@ -16,7 +16,7 @@ type SnippetMatch = {
   readonly matchLevel: 'none' | 'partial' | 'full';
 };
 
-type SnippetResult<THit> = THit extends string | number
+export type SnippetResult<THit> = THit extends string | number
   ? SnippetMatch
   : {
       [KAttribute in keyof THit]: SnippetResult<THit[KAttribute]>;
--- a/packages/client-search/src/types/Hit.ts
+++ b/packages/client-search/src/types/Hit.ts
@@ -1,0 +1,47 @@
+type HighlightMatch = {
+  readonly value: string;
+  readonly matchLevel: 'none' | 'partial' | 'full';
+  readonly matchedWords: readonly string[];
+  readonly fullyHighlighted?: boolean;
+};
+
+type Highlight<THit> = THit extends string | number
+  ? HighlightMatch
+  : {
+      [KAttribute in keyof THit]: Highlight<THit[KAttribute]>;
+    };
+
+type SnippetMatch = {
+  readonly value: string;
+  readonly matchLevel: 'none' | 'partial' | 'full';
+};
+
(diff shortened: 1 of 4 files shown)


packages/client-search/src/methods/index/getObject.ts
```diff
@@ -2,13 +2,13 @@ import { encode } from '@algolia/client-common';
...
 
-import { GetObjectOptions, ObjectWithObjectID, SearchIndex } from '../..';
+import { GetObjectOptions, Hit, SearchIndex } from '../..';
 
 export const getObject = (base: SearchIndex) => {
   return <TObject>(
     objectID: string,
     requestOptions?: RequestOptions & GetObjectOptions
-  ): Readonly<Promise<TObject & ObjectWithObjectID>> => {
+  ): Readonly<Promise<Hit<TObject>>> => {
```
Haroenv, line 11: there's no highlighting on getObject I think

packages/client-search/src/types/BrowseOptions.ts
```diff
@@ -1,11 +1,11 @@
 import { BrowseResponse } from '.';
-import { ObjectWithObjectID } from './ObjectWithObjectID';
+import { Hit } from './Hit';
 
 export type BrowseOptions<TObject> = {
   /**
    * The callback called for each batch of objects.
    */
-  readonly batch?: (batch: ReadonlyArray<TObject & ObjectWithObjectID>) => any;
+  readonly batch?: (batch: ReadonlyArray<Hit<TObject>>) => any;
```
Haroenv, line 8: there's no highlighting on browse either, but I think e.g. _rankingInfo can be retrieved

Haroenv (commented):
looks good, but I think you used it in too many places, getObject(s) & browse don't have highlighting

packages/client-search/src/types/Hit.ts
```diff
@@ -0,0 +1,47 @@
...
+type Snippet<THit> = THit extends string | number
+  ? SnippetMatch
+  : {
+      [KAttribute in keyof THit]: Snippet<THit[KAttribute]>;
+    };
+
+export type Hit<THit> = THit & {
+  readonly objectID: string;
+  readonly _highlightResult?: Highlight<THit>;
+  readonly _snippetResult?: Snippet<THit>;
```
Haroenv, line 28: I'm not sure, but maybe change the name of the type to HighlightResult and export it?

packages/client-search/src/methods/index/getObject.ts
francoischalifour: That's right indeed. I'm just gonna change the `*Response` types.

packages/client-search/src/types/Hit.ts
francoischalifour: See f3efb64080ded7a67aecf8b576523e84595c4906 and 82d6f8e98ddb2987daf6ca8b8db829b988dd5f12.

packages/client-search/src/types/BrowseOptions.ts
francoischalifour: See fc268e0eee87b835fe85eeaf0d9225c31e9d5b35.

packages/client-search/src/methods/index/getObject.ts
francoischalifour: See fc268e0eee87b835fe85eeaf0d9225c31e9d5b35.

Haroenv (approved):
looks very good!

In the future I'd love to play with `attributesToRetrieve` & `attributesToHighlight` deciding which keys get responded, although I'm not sure if you can get all of that statically (since it could be a setting)

PR #1210, opened 2020-09-28
chore: migrate release process to Ship.js

aseure:
chore: migrate release process to Ship.js

This commit introduces a new release process by:

- removing the need for the scripts/release.sh bash script in favor of
  https://github.com/algolia/shipjs.
- automating the release in CircleCI

Notable changes are:

- `version` attribute added to the `package.json` file: this is a bit
  redundant with the `version` attribute available in the `lerna.json`
  but it is required by `conventional-changelog-cli`, which is used by Ship.js,
  to generate new Changelog entries, and can only identify versions based
  on the `package.json` file.
- `CHANGELOG.md` text header has been removed because
  `conventional-changelog-ci` is not aware of headers in Changelog
  files.
- `GITHUB_TOKEN` environment variable is now added to the CircleCI
  configuration. It grants CircleCI with the appropriate rights to run
  `shipjs trigger`, effectively publishing to NPM. The value is set to
  the Github account `algolia-api-client-bot` which can be retrieved
  from our Vault cluster, like other credentials.
- `CONTRIBUTING.md` file updated to reflect the updated release process.


ship.config.js
```diff
@@ -0,0 +1,30 @@
...
+module.exports = {
+  monorepo: {
+    mainVersionFile: 'package.json',
+    packagesToBump: ['packages/*'],
+    packagesToPublish: ['packages/*'],
+  },
+  getTagName: ({ version }) => `${version}`,
+  conventionalChangelogArgs:
+    '--config conventional-changelog.config.js --infile CHANGELOG.md --same-file',
+  versionUpdated({ version, dir }) {
```
eunjae-lee, line 16: You're gonna need to update the version of the the root `package.json` here.

eunjae-lee:
You may want to address in `CONTIRBUTING.md` about how you should prepend `GITHUB_TOKEN` or add it to `.env` file before running `npm run release`.

ship.config.js
aseure: This is already done by the `monerepo.mainVersionFile` set to `package.json` on line 9, along with the update of others package-specific `package.json` files.

Here in `versionUpdated`, I'm only manually updating the root `lerna.json` and `packages/client-common/src/version.ts` TypeScript file which aren't done automatically.

aseure:
@eunjae-lee thanks for the review buddy and for your help with Ship.js! I've addressed your comments.

> You may want to address in CONTIRBUTING.md about how you should prepend GITHUB_TOKEN or add it to .env file before running npm run release.

Good point, just added it to the CONTRIBUTING.md file 👍

eunjae-lee (dismissed):
The change looks good to me, but I see a circleci job failing and some other jobs are not triggered yet.

aseure:
@Haroenv @eunjae-lee The `Required` checks are pending because I've had to explicitely rename the CircleCI jobs for this to pass. Are you fine if I edit the branch protection rules on my own right now or do you prefer to do it on your own later (because they are some pending pull requests opened as well).

I've updated the Confluence documentation and updated the CONTRIBUTING.md to reflect this, added our credentials to Vault. Could you have a last review before I merge this? Thanks!

.circleci/config.yml
```diff
@@ -119,27 +119,94 @@ jobs:
...
+            - test_unit_12
+            - test_unit_13
+            - test_unit_14
+            - test_lint
+            - test_types
+            - test_build
+            - test_browser
+          filters:
+            tags:
+              only: /^[1-9]+.[0-9]+.[0-9]+.*/
```
Haroenv, line 210: doesn't it this way still only run for tags?

eunjae-lee: https://github.com/algolia/algoliasearch-client-javascript/pull/1210/commits/5786f1afaf4f6ac63a04e1a6918c6aa883c3254f

aseure: Yes my bad, I've left the filters we are using on all other API clients but in your case, you only need to build on `master`, since `shipjs trigger` already filters based on the commit message.

Perhaps the other filters are not needed anymore in that case (in all dependent jobs).

.circleci/config.yml
```diff
@@ -119,27 +119,93 @@ jobs:
...
+            yarn shipjs trigger
+
 workflows:
   version: 2
   ci:
     jobs:
       - test_unit:
           version: '8'
+          name: 'test_unit_8'
+          filters:
```
Haroenv, line 143: I think most of these filters can also be removed

eunjae-lee: https://github.com/algolia/algoliasearch-client-javascript/pull/1210/commits/3198d231ec0bb705f8d1d4f1b2df95aebf87d2e9

Haroenv (approved):
this looks good to me!

eunjae-lee:
@Haroenv Do you know what's going on here?
image

The job names seem to be messed up.

Haroenv:
the names of the tests have changed, so it's expected the "required" ones aren't ran. After we merge this PR we need to change which ones are required

eunjae-lee:
> the names of the tests have changed, so it's expected the "required" ones aren't ran. After we merge this PR we need to change which ones are required

Ah I didn't know the names have changed in this PR.

PR #1213, opened 2020-09-29
feat(client-search): add personalization field to RankingInfo

kombucha:
- Extract and expose `RankingInfo` type
- Add missing personalization field for ranking info


Haroenv (approved):
thanks!

PR #1215, opened 2020-10-18
fix(node): allow passing requesterOptions

bodinsamuel:
On the crawler we are trying to control more our http connections 
and we would like to modify the behavior of the client to test a few things, especially the `family` options.

This option is related to ip resolution and a "fix" to reduse some network errors like: `ENOTFOUND` `EAI_AGAIN`
Those errors are the most common ones that triggers the infamous "Unreachable hosts"
ref: https://github.com/nodejs/node/issues/5436#issuecomment-189474356
(personal opinion would be to use `family: 4` by default, DM for more about this)

If you are not confident letting people modify the whole options, we'll be fine with just the `family` option.

### DIFF
--- a/packages/requester-node-http/src/__tests__/unit/node-http-requester.test.ts
+++ b/packages/requester-node-http/src/__tests__/unit/node-http-requester.test.ts
@@ -251,3 +251,31 @@ describe('error handling', (): void => {
     expect(response.isTimedOut).toBe(false);
   });
 });
+
+describe('requesterOptions', () => {
+  it('allows to pass requesterOptions', async () => {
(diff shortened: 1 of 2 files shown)


Haroenv:
I think it works better like this: https://github.com/algolia/algoliasearch-client-javascript/pull/1184, I just closed the PR since my use case was for an unofficial API and I didn't see the use case other than my own.

Do you think you could take an approach like that PR?

bodinsamuel:
That means, I would need to pass the request options to all call?

Haroenv (approved):

bodinsamuel:
thanks @Haroenv 🙏🏻

PR #1218, opened 2020-11-02
chore(release): configure shipjs to use lerna

Haroenv:
basically if you don't use lerna to update the versions, internal versions get out of sync and the dependencies never get updated (even if we would run lerna version).

This PR solves that by preventing shipjs updating the versions and doing it all with lerna in versionUpdated hook


eunjae-lee (approved):
looks good

PR #1223, opened 2020-11-10
feat(analytics-reco): allow creation of clients with their own credentials

Haroenv:
see https://github.com/algolia/algoliasearch-client-specs-internal/issues/144

This actually didn't need to change any code, just the types, since they didn't allow appId & apiKey, while they actually are passed along already

### DIFF
--- a/packages/algoliasearch/src/types/AlgoliaSearchOptions.ts
+++ b/packages/algoliasearch/src/types/AlgoliaSearchOptions.ts
@@ -3,13 +3,22 @@ import { ClientTransporterOptions } from '@algolia/client-common';
 import { RecommendationClientOptions } from '@algolia/client-recommendation';
 import { SearchClientOptions } from '@algolia/client-search';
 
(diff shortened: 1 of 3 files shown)


packages/algoliasearch/src/types/AlgoliaSearchOptions.ts
```diff
@@ -3,13 +3,22 @@ import { ClientTransporterOptions } from '@algolia/client-common';
...
+type Credentials = { readonly appId: string; readonly apiKey: string };
+export type WithoutCredentials<TClientOptions extends Credentials> = Omit<
+  TClientOptions,
+  keyof Credentials
+>;
+export type OptionalCredentials<TClientOptions extends Credentials> = Omit<
+  TClientOptions,
+  keyof Credentials
+> &
+  Pick<Partial<TClientOptions>, keyof Credentials>;
```
Haroenv, line 15: if someone can find a cleaner way to implement this, I'm all ears

eunjae-lee (approved):
Looks good to me. I think the typing is okay.

PR #1227, opened 2020-11-18
fix(findAnswers): expose in lite version

eunjae-lee:
## Summary

This PR exposes `findAnswers` method in the lite version. In our documents, we kind of encourage to use `algoliasearch/lite` a lot. So we'd better have this method in the lite version as well for better adoption.

### DIFF
--- a/packages/algoliasearch/src/builds/browserLite.ts
+++ b/packages/algoliasearch/src/builds/browserLite.ts
@@ -4,6 +4,9 @@ import { createInMemoryCache } from '@algolia/cache-in-memory';
 import { AuthMode, version } from '@algolia/client-common';
 import {
   createSearchClient,
+  findAnswers,
+  FindAnswersOptions,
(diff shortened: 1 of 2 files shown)


Haroenv:
I didn't do this for bundlesize, but I guess it makes sense. You still need to update the size :)

eunjae-lee:
> I didn't do this for bundlesize, but I guess it makes sense. You still need to update the size :)

It didn't exceed the current threshold :)

Haroenv:
can you add the integration test on browser-lite now too? it's https://github.com/algolia/algoliasearch-client-javascript/blob/master/specs/answers.spec.ts (you can see https://github.com/algolia/algoliasearch-client-javascript/blob/master/specs/search.spec.ts for lite usage)

Haroenv (approved):
lgtm once tests pass 👍

PR #1229, opened 2020-11-19
fix(ts): correct search response for facets_stats

Haroenv:
https://www.algolia.com/doc/api-reference/api-methods/search/#method-response-facets_stats

as far as I can tell this was always wrong in v4

### DIFF
--- a/packages/client-search/src/types/SearchResponse.ts
+++ b/packages/client-search/src/types/SearchResponse.ts
@@ -67,7 +67,7 @@ export type SearchResponse<TObject = {}> = {
   /**
    * Statistics for numerical facets.
    */
-  facetsStats?: Record<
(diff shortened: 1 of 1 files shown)


eunjae-lee:
Huh weirdly only key not in camelCase 
https://www.algolia.com/doc/api-reference/api-methods/search/#method-response-facets_stats

eunjae-lee (approved):

nojvek:
Just want say, thanks for fixing this. This has led me to at-least five `// @ts-expect-error` comments within the codebase.

Haroenv:
Thanks @nojvek, if there's other places where you're currently suppressing errors, feel free to open pull requests / issues for that!

PR #1234, opened 2020-11-25
fix(ts): don't require every attribute to be highlighted

Haroenv:
we didn't notice this earlier, since in InstantSearch.js we didn't yet update to the version of algoliasearch including the highlighting. The type in algoliasearch is better than the one in InstantSearch though, so we can find a way to switch in the future


francoischalifour (approved):
Nice

PR #1238, opened 2020-12-10
Update multipleQueries.ts

Gregoirevda:
I would like to bypass the cache for the same request. (eg refreshing) I'm not sure if there's another way to do this? (passing `must-revalidate` cache-control header for example)

is `cacheable` for request cache only or for both request and response?

### DIFF
--- a/packages/client-search/src/methods/client/multipleQueries.ts
+++ b/packages/client-search/src/methods/client/multipleQueries.ts
@@ -27,7 +27,7 @@ export const multipleQueries = (base: SearchClient) => {
         data: {
           requests,
         },
-        cacheable: true,
+        cacheable: (requestOptions && requestOptions.cacheable) || true,
       },
       requestOptions
     );


Gregoirevda:
I've tested passing headers, but the `cacheable` option has priority over the actual request being sent. So this doesn't work
```js
index.search('', {
  headers: {
     'Cache-Control': 'no-cache'
  }
})
```

Passing `cacheable` as false works, but unfortunately it bypasses both request and response cache.
```js
index.search('', {
  cacheable: false
})
```

What's the best way to only bypass the request cache? eg: mobile pull to refresh

Haroenv:
You likely want to prevent both the request & response cache if you want a fresh response completely. In general we go for a `.clearCache()` then `.search()` again with the same parameters. In InstantSearch this is implemented with a `refresh` function. 

As you noticed, you can already override the cache by passing `{cacheable: false}` to the request options

The cache is purely client-side, so headers indeed do nothing.

Haroenv:
One thing that might have been confusing is that the cache is talking about the search client cache, nothing to do with the search engine (which doesn't cache)

As there are already ways to clear the client's cache, I'll close this pull request, but sorry for losing track and not replying earlier

PR #1240, opened 2021-01-14
feat(ts): add decompoundQuery

Haroenv:
A new setting and parameter as an add-on to [decompoundedAttributes](https://www.algolia.com/doc/api-reference/api-parameters/decompoundedAttributes/?client=javascript)

see https://github.com/algolia/algoliasearch-client-specs-internal/issues/153
fixes https://github.com/algolia/algoliasearch-client-javascript/issues/1242

### DIFF
--- a/packages/client-search/src/types/SearchOptions.ts
+++ b/packages/client-search/src/types/SearchOptions.ts
@@ -334,4 +334,12 @@ export type SearchOptions = {
    * Engine's default: true
    */
   readonly enableABTest?: boolean;
+
+  /**
+   * Enable word segmentation (also called decompounding) at query time for
+   * languages that have it enabled using "decompoundedAttributes". For example,
+   * this turns the query "spaanplaatbehang" (Dutch) into "spaan plaat behang"
+   * to retrieve more relevant results.
+   */
+  readonly decompoundQuery?: boolean;
 };
--- a/packages/client-search/src/types/SearchOptions.ts
+++ b/packages/client-search/src/types/SearchOptions.ts
@@ -337,9 +337,9 @@ export type SearchOptions = {
 
   /**
    * Enable word segmentation (also called decompounding) at query time for
-   * languages that have it enabled using "decompoundedAttributes". For example,
-   * this turns the query "spaanplaatbehang" (Dutch) into "spaan plaat behang"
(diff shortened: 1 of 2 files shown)


packages/client-search/src/types/SearchOptions.ts
```diff
@@ -334,4 +334,12 @@ export type SearchOptions = {
    * Engine's default: true
    */
   readonly enableABTest?: boolean;
+
+  /**
+   * Enable word segmentation (also called decompounding) at query time for
+   * languages that have it enabled using "decompoundedAttributes". For example,
```
remy-zeiss, line 340:
```suggestion
   * compatible languages (Danish, Dutch, Finnish, German, Norwegian, Swedish) . For example,
```

remy-zeiss: This setting is not linked to decompoundedAttributes, beyond its behaviour. You can use one or the other only independantly (that would be weird, but possible).

packages/client-search/src/types/Settings.ts
```diff
@@ -283,4 +283,12 @@ export type Settings = {
    * Custom userData that could be added to the Settings.
    */
   readonly userData?: any;
+
+  /**
+   * Enable word segmentation (also called decompounding) at query time for
+   * languages that have it enabled using "decompoundedAttributes". For example,
```
remy-zeiss, line 289:
```suggestion
   * compatible languages (Danish, Dutch, Finnish, German, Norwegian, Swedish) . For example,
```

packages/client-search/src/types/SearchOptions.ts
Haroenv: is this list never going to change? I'd rather have the website be the source of truth for the languages list

remy-zeiss: It might, but not in the foreseeable future. 
That being said you make a good point, so I'll leave it up to you to decide. :)

Haroenv: I removed the mention of decompoundedAttributes for now, let me know what you think

remy-zeiss (approved):
Perfect !

chloelbn (approved):

Issue #1243, opened 2021-01-15
feat(settings): add attributesToTransliterate settings


chloelbn:
Add a new index settings called attributesToTransliterate. This setting can take a a list of attributes or wildcard *

PR #1244, opened 2021-01-15
feat(ts): add the attributesToTransliterate setting

Haroenv:
fixes https://github.com/algolia/algoliasearch-client-javascript/issues/1243

### DIFF
--- a/packages/client-search/src/types/Settings.ts
+++ b/packages/client-search/src/types/Settings.ts
@@ -291,4 +291,10 @@ export type Settings = {
    * results.
    */
   readonly decompoundQuery?: boolean;
+
+  /**
+   * Specify on which attributes in your index Algolia should apply
+   * transliteration to a normalized form for compatible scripts.
+   */
+  readonly attributesToTransliterate?: readonly string[];
 };
--- a/packages/client-search/src/types/Settings.ts
+++ b/packages/client-search/src/types/Settings.ts
@@ -294,7 +294,7 @@ export type Settings = {
 
   /**
    * Specify on which attributes in your index Algolia should apply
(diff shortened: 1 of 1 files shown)


chloelbn:
Note: this attribute is only applied when your query language is Japanese

chloelbn (dismissed):
Note: this attribute is only applied when your query language is Japanese

Haroenv:
doesn't "for compatible scripts" imply that already? @chloelbn

chloelbn:
@Haroenv indeed! I didn't get that with the wording sorry.

chloelbn (approved):

packages/client-search/src/types/Settings.ts
```diff
@@ -291,4 +291,10 @@ export type Settings = {
    * results.
    */
   readonly decompoundQuery?: boolean;
+
+  /**
+   * Specify on which attributes in your index Algolia should apply
+   * transliteration to a normalized form for compatible scripts.
```
remy-zeiss, line 297:
```suggestion
   * Specify on which attributes in your index Algolia Japanese transliteration should apply
   * to make those attributes searchable in Hiragana.
```

remy-zeiss, line 297:
```suggestion
   * Specify on which attributes in your index Algolia Japanese transliteration should apply
   * to make those attributes searchable in Hiragana.
```

remy-zeiss, line 297:
```suggestion
   * Specify on which attributes in your index Algolia should apply
   * Japanese transliteration to make words indexed in Katakana or Kanji searchable in Hiragana.
```

PR #1253, opened 2021-02-17
feat(dictionaries): adds methods and tests

### LINKED ISSUE #1248: (feat): Custom Dictionaries
Implement custom dictionaries methods
See [Ruby PR](https://github.com/algolia/algoliasearch-client-ruby/pull/445) as an example.

chloelbn:
Fixes #1248


packages/algoliasearch/src/builds/node.ts
```diff
@@ -609,6 +635,41 @@ export type SearchClient = BaseSearchClient & {
...
+    requestOptions?: RequestOptions
+  ) => Readonly<Promise<GetDictionarySettingsResponse>>;
+  readonly setDictionarySettings: (
+    settings: readonly DictionarySettings[],
+    requestOptions?: RequestOptions
+  ) => Readonly<WaitablePromise<SetDictionarySettingsResponse>>;
+  readonly getDictionaryTask: (
+    taskID: number,
+    requestOptions?: RequestOptions
+  ) => Readonly<Promise<TaskStatusResponse>>;
```
Haroenv, line 672: what's this method for? should it not be a generic waitTask?

packages/client-common/src/__tests__/TestSuite.ts
```diff
@@ -68,6 +68,13 @@ export class TestSuite {
     return client;
   }
 
+  public makeDictionarySearchClient(
+    appIdEnv: string = 'ALGOLIA_APPLICATION_ID_2',
+    apiKeyEnv: string = 'ALGOLIA_ADMIN_KEY_2'
+  ) {
+    return this.makeSearchClient(appIdEnv, apiKeyEnv);
```
Haroenv, line 75: this method isn't needed, we can use makeSearchClient in the test with these credentials instead :)

packages/algoliasearch/src/builds/node.ts
chloelbn: it should indeed but I had issues with the fact there were two methods named the same way so I renamed it. It should however be changed for the generic waitTask indeed

Haroenv:
I wonder if the tests are failing because we are doing account tasks in parallel, what do you think @chloelbn?

packages/client-search/src/types/GetDictionarySettingsResponse.ts
```diff
@@ -0,0 +1,11 @@
+import { DictionaryName } from './DictionaryName';
+import { RequireAtLeastOne } from './RequireAtLeastOne';
+
+export type GetDictionarySettingsResponse = {
+  /**
+   * TODO: Description
```
Haroenv, line 6:
```suggestion
   * Disable the builtin Algolia entries for a type of dictionary per language
```

something like that?

packages/client-search/src/types/SaveDictionaryEntriesOptions.ts
```diff
@@ -0,0 +1,8 @@
+import { DictionaryEntry } from './DictionaryEntry';
+
+export type SaveDictionaryEntriesOptions = {
```
Haroenv, line 3: I think this should be called something like 

```suggestion
export type DictionaryEntriesOptions = {
```

so that it doesn't have the `save` part in it, it's used in more params

Haroenv (commented):
looks great, thanks @shortcuts for finalising this!

packages/client-search/src/types/GetDictionarySettingsResponse.ts
shortcuts: Oops I forgot about these, make sense!

packages/client-search/src/types/SaveDictionaryEntriesOptions.ts
shortcuts: Definitely!

packages/client-search/src/__tests__/integration/dictionary.test.ts
```diff
@@ -32,8 +32,13 @@ describe(testSuite.testName, () => {
     const stopwords = await client.searchDictionaryEntries('stopwords', stopwordEntry.objectID);
 
     expect(stopwords.nbHits).toEqual(nbSearchEntries + 1);
-    expect(stopwords.hits[nbSearchEntries].objectID).toEqual(stopwordEntry.objectID);
-    expect(stopwords.hits[nbSearchEntries].word).toEqual(stopwordEntry.word);
+    expect(stopwords.hits).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          ...stopwordEntry,
+        }),
```
Haroenv, line 39:
```suggestion
        expect.objectContaining(stopwordEntry),
```

this is essentially the same

Haroenv (approved):
lgtm on green!

PR #1254, opened 2021-02-26
Spell success properly

greg0ire:
Warning: I'm not a JS developer so I have no idea if there might be BC-breaks in this PR or not. I also don't know if this might actually fix any bug, hence why I picked `refactor` here.

### DIFF
--- a/packages/transporter/src/concerns/retryDecision.ts
+++ b/packages/transporter/src/concerns/retryDecision.ts
@@ -26,7 +26,7 @@ export const retryDecision = <TResponse>(
   }
 
   if (isSuccess(response)) {
-    return outcomes.onSucess(response);
+    return outcomes.onSuccess(response);
   }
 
   return outcomes.onFail(response);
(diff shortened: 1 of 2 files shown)


Haroenv:
hi @greg0ire, thanks a lot for this PR, but as you guessed it, this would be a breaking change for those who write a custom requester today. It would be possible to alias it to the correct name, and accept both wrong and correct spelling, or to add a comment `// @MAJOR: replace with onSuccess` so this can be changed in a next major version

greg0ire:
> It would be possible to alias it to the correct name, and accept both wrong and correct spelling

Is there a standard way in JS to trigger a deprecation when a call is performed with the wrong spelling?

Haroenv:
No, there's no standard way of that, it would be calling the previous function and adding a log. At the moment for the sake of avoiding a major change for just this, I'd be edging on the side of having just a comment for now

Haroenv (approved):

PR #1256, opened 2021-03-03
fix(answers): rename params to searchParameters

eunjae-lee:
# Summary

This PR renames `params` to `searchParameters` in `FindAnswersOptions`.

`params` is still supported but the new `searchParameters` is recommended for better clarification.

Should we leave `params` and mark it as deprecated? Or just delete it since it's a type and deleting it won't affect runtime?


Haroenv (approved):
it's ok like this, since either version works in javascript, can you make a PR to the search client with the same too?

shortcuts (approved):

PR #1270, opened 2021-04-28
fix(api-keys): fix typing issues

kombucha:
- `createdAt` is a unix timestamp when using `listApiKeys` or `getApiKey`, not a `string`
- `UpdateApiKeyOptions` is missing the `acl` attribute (See [documentation](https://www.algolia.com/doc/api-reference/api-methods/update-api-key/#method-param-acl))


Haroenv (approved):
thanks!

PR #1273, opened 2021-06-10
feat(ts): document renderingContent

Haroenv:
DX-2066

see also the helper PR: https://github.com/algolia/algoliasearch-helper-js/pull/823

### DIFF
--- a/packages/client-search/src/types/Rule.ts
+++ b/packages/client-search/src/types/Rule.ts
@@ -1,4 +1,4 @@
-import { SearchOptions } from '.';
+import { SearchOptions, Settings } from '.';
 
 export type Rule = {
   /**
@@ -125,6 +125,12 @@ export type ConsequenceParams = {
    * Behaves like optionalFilters.
    */
   readonly automaticOptionalFacetFilters?: readonly AutomaticFacetFilter[];
+
+  /**
+   * Content defining how the search interface should be rendered.
+   * A default value for this can be set via settings
+   */
+  readonly renderingContent?: Settings['renderingContent'];
 };
 
 export type Condition = {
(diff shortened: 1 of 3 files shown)


packages/client-search/src/types/SearchResponse.ts
```diff
@@ -225,4 +225,6 @@ export type SearchResponse<TObject = {}> = {
    * The relevancy threshold applied to search in a virtual index.
    */
   appliedRelevancyStrictness?: number;
+
+  readonly renderingContent?: Settings['renderingContent'];
```
eunjae-lee, line 229: Why is it `readonly` here? I feel like technically all the fields in this response is readonly, though.

Haroenv: ah yes it shouldn't be, copied from the other place

Haroenv:
```suggestion
  renderingContent?: Settings['renderingContent'];
```

eunjae-lee (approved):

packages/client-search/src/types/Rule.ts
```diff
@@ -125,6 +125,12 @@ export type ConsequenceParams = {
    * Behaves like optionalFilters.
    */
   readonly automaticOptionalFacetFilters?: readonly AutomaticFacetFilter[];
+
+  /**
+   * Content defining how the search interface should be rendered.
+   * A default value for this can be set via settings
+   */
+  readonly renderingContent?: Settings['renderingContent'];
```
VladislavFitz, line 133: I probably missed it, but did we agree on `renderingContent` as a part of `Rule.Consequence`?

Haroenv: In this object it's part of consequence.params.renderingContent, as we discussed

VladislavFitz (approved):

PR #1278, opened 2021-06-17
feat(personalization): deprecate client-recommendation

francoischalifour:
_This PR targets the `feat/recommend` base branch._

## Description

We want to remove mentions of the old Recommendation API to avoid any confusion with the new Algolia Recommend product.

This deprecates the `@algolia/client-recommendation` package and migrates it to `@algolia/client-personalization`.

## Changes

- Deprecate `client-recommendation` exports in TSDoc
- Add a depreciation warning when `createRecommendationClient` is called
- Update the endpoint from `recommendation.us.algolia.com` to `personalization.us.algolia.com`
- Add `initPersonalization` to `algoliasearch` and deprecate `initRecommendation`

fixes https://github.com/algolia/algoliasearch-client-javascript/issues/1281

## What's next

- Deprecate the npm package `@algolia/client-recommendation` in the Recommend release


packages/algoliasearch/src/builds/browser.ts
```diff
@@ -319,10 +319,26 @@ export default function algoliasearch(
...
+        return createPersonalizationClient({
+          ...commonOptions,
+          ...clientOptions,
+          methods: {
+            getPersonalizationStrategy,
+            setPersonalizationStrategy,
+          },
+        });
+      },
       initRecommendation: () => (
```
Haroenv, line 334: does this work like this?
```suggestion
      initRecommendation: (args)  => {warn(...) return this.initPersonalization(args)}
```

(we don't have a `deprecate` in this codebase at the moment I think, it would fit nicely otherwise

Haroenv (commented):
looks great! one place where it can possible be a bit simpler maybe?

packages/algoliasearch/src/builds/browser.ts
francoischalifour: Not really but I went with 95cf1089c7af9dbc2772c74d89357ce2d1367608.

packages/client-recommendation/package.json
```diff
@@ -17,7 +17,7 @@
   ],
   "dependencies": {
     "@algolia/client-common": "4.9.3",
-    "@algolia/requester-common": "4.9.3",
-    "@algolia/transporter": "4.9.3"
+    "@algolia/client-personalization": "4.9.3",
+    "@algolia/requester-common": "4.9.3"
```
Haroenv, line 21:
```suggestion
```

I don't think this is still used here. Can you double-check?

packages/client-search/src/types/ApiKeyType.ts
```diff
@@ -7,6 +7,7 @@ export const ApiKeyACLEnum: Readonly<Record<string, ApiKeyACLType>> = {
   EditSettings: 'editSettings',
   ListIndexes: 'listIndexes',
   Logs: 'logs',
+  Personalization: 'personalization',
   Recommendation: 'recommendation',
```
Haroenv, line 11: is one of these two deprecated?

packages/client-recommendation/package.json
francoischalifour: It is: https://github.com/algolia/algoliasearch-client-javascript/blob/95cf1089c7af9dbc2772c74d89357ce2d1367608/packages/client-recommendation/src/methods/setPersonalizationStrategy.ts#L1

packages/client-search/src/types/ApiKeyType.ts
francoischalifour: I guess the two will work for some time. Wdym?

packages/client-recommendation/package.json
Haroenv: ah yes good call! I was looking in the wrong files

packages/client-search/src/types/ApiKeyType.ts
Haroenv: good enough for now!

Haroenv (approved):

PR #1279, opened 2021-06-20
fix(Rule): automaticOptionalFacetFilters take the same input as automaticFacetFilters

bodinsamuel:
Unless the documentation is wrong, the type should be the same.
(that also means there is no test for this use case)


Haroenv (approved):

Haroenv:
thanks @bodinsamuel !

PR #1280, opened 2021-06-21
feat(recommend): introduce Recommend API client

francoischalifour:
This adds a new API client for Recommend.

## Usage

```ts
import algoliarecommend from '@algolia/recommend';

const recommendClient = algoliarecommend('appId', 'apiKey');

recommendClient
  .getRelatedProducts({
    /* ... */
  })
  .then((response) => {
    console.log(response);
  });
```

### `getRelatedProducts`

Returns items from the `related-products` AI model.

### `getFrequentlyBoughtTogether`

Returns items from the `bought-together` AI model.

### `getRecommendations`

Returns items from the provided AI model (other methods are based on that one).

## Next steps

- Add documentation on the Algolia website


redox:
> ```ts
> recommend
> ```

Shouldn't we rather us `algoliarecommend` like we do for `algoliasearch`?

Haroenv:
algoliarecommend is quite hard to read, don't you think @redox ?

packages/recommend/src/__tests__/getFrequentlyBoughtTogether.test.ts
```diff
@@ -0,0 +1,83 @@
...
+describe('getFrequentlyBoughtTogether', () => {
+  test('builds the request', async () => {
+    const client = createMockedClient();
+
+    await client.getFrequentlyBoughtTogether(
+      {
+        indexName: 'products',
+        objectID: 'B018APC4LE',
+      },
+      {}
```
Haroenv, line 25: is this a leftover from providing custom hosts?

packages/recommend/src/__tests__/getRecommendations.test.ts
```diff
@@ -0,0 +1,84 @@
+import { TestSuite } from '../../../client-common/src/__tests__/TestSuite';
+
+const recommend = new TestSuite('recommend').recommend;
+
+function createMockedClient() {
+  const client = recommend('appId', 'apiKey');
+  // @ts-ignore read-only property
+  client.transporter = {
+    ...client.transporter,
+    read: jest.fn(),
```
Haroenv, line 10: can this be done with jest.spyOn instead to avoid the mutation? not a big deal though

packages/recommend/src/createUserAgent.ts
```diff
@@ -0,0 +1,23 @@
+import { UserAgent, UserAgentOptions } from '@algolia/transporter';
+
+// Copied from https://github.com/algolia/algoliasearch-client-javascript/blob/831a3debd0c19c87fd71a3e09396bcb3eacd1f5c/packages/transporter/src/createUserAgent.ts
+// and adapted the initial value for Recommend.
+export function createUserAgent(version: string): UserAgent {
+  const userAgent: UserAgent = {
+    value: `Algolia Recommend for JavaScript (${version})`,
```
Haroenv, line 7: maybe the default should be:
```suggestion
    value: `Algolia for JavaScript (${version}); Recommend (${version})`,
```

not 100% sure, but this would avoid being marked as an unknown client

packages/recommend/src/methods/getRecommendations.ts
```diff
@@ -0,0 +1,47 @@
...
+type GetRecommendations = (
+  base: RecommendClient
+) => WithRecommendMethods<RecommendClient>['getRecommendations'];
+
+export const getRecommendations: GetRecommendations = base => {
+  return (options, requestOptions) => {
+    const requests: readonly GetRecommendationsOptions[] = [
+      {
+        // @TODO: we temporarly need to provide `queryParameters` because the
+        // endpoint doesn't default it yet.
```
Haroenv, line 25: could we add a ticket number / GitHub issue here? or is there only private info available?

Haroenv (commented):
looks good overall, some small questions. Agreed with Sylvain that we need to be 100% sure of the client name before publishing

redox:
> algoliarecommend is quite hard to read, don't you think @redox ?

I think this is not worst than `algoliasearch` (we are just less used to it), and I thought we should rather keep consistency here.

packages/recommend/src/createUserAgent.ts
redox: Agreed, we don't want to differentiate the API client when used for Search vs Recommend. I think we should rather use the `addAlgoliaAgent` to amend the regular default one.

francoischalifour: Thanks for the rectification. I thought we wanted to differentiate them because they are two different products.

francoischalifour: See 827022f32f623ce460cb36df593472e5e233d28f.

packages/recommend/src/__tests__/getRecommendations.test.ts
francoischalifour: Good point—see 9d919a63dbb71d38486b3eab7ff2c724a7de36c3.

packages/recommend/src/__tests__/getFrequentlyBoughtTogether.test.ts
francoischalifour: Nope, it's to make sure that the `requestOptions` are forwarded.

packages/recommend/src/methods/getRecommendations.ts
francoischalifour: It's more that the server crashes because the engine assume that `queryParameters` always exists. The engine team told me them they were fixing this asap. I'm in touch with them to remove this line of code as soon as it's not necessary anymore.

francoischalifour:
@redox Is your reasoning that "specified" clients ready to be used shouldn't be imported with the `@algolia` npm namespace, but "productized" with `algolia*`?

(I'm fine with both naming.)

packages/recommend/src/methods/getRecommendations.ts
francoischalifour: It's fixed! See 8c8c3030eeab0691686f40387e0dbf3436874e35.

redox:
> @redox Is your reasoning that "specified" clients ready to be used shouldn't be imported with the `@algolia` npm namespace, but "productized" with `algolia*`?
> 
> (I'm fine with both naming.)

I like the new `@algolia/recommend` package naming (instead of `algoliarecommend` or `algoliasearch`), my point was more about the name of the exported function being `recommend` VS `algoliarecommend`. The same way we don't use `search` but `algoliasearch`.

francoischalifour:
@redox Right, the exported function isn't named (it's a default export, users can name it the way they want), so it's just a matter of documentation.

We can document the usage this way:

```ts
import algoliarecommend from '@algolia/recommend';

const recommendClient = algoliarecommend('appId', 'apiKey');
```

Wdyt?

Haroenv (commented):
we just need to add a readme, but ☑️ otherwise!

Haroenv (approved):
good call fixing that link!

packages/recommend/src/methods/getRecommendations.ts
```diff
@@ -0,0 +1,44 @@
...
+import { RecommendClient, RecommendModel, WithRecommendMethods } from '../types';
+
+export type GetRecommendationsOptions = {
+  readonly indexName: string;
+  readonly model: RecommendModel;
+  readonly objectID: string;
+  readonly threshold?: number;
+  readonly maxRecommendations?: number;
+  readonly queryParameters?: SearchOptions;
+  readonly fallbackParameters?: Pick<SearchOptions, 'facetFilters'>;
```
shortcuts, line 13: Shouldn't they both have the same type?

```ts
type GetRecommendationsSearchParameters  = Omit<
  SearchOptions,
  'page' | 'hitsPerPage' | 'offset' | 'length'
>;
```

shortcuts (approved):
Small question but looks good overall!

packages/recommend/src/methods/getRecommendations.ts
francoischalifour: Good call! See 911c7db50f125738fe741b11591cc0e64d4e67a8.

shortcuts (dismissed):
Wonderful

shortcuts (approved):

redox:
👏 well done team!

PR #1283, opened 2021-06-23
feat(recommend): support multiple queries

francoischalifour:
## Description

Before releasing the new Recommend API client, we changed the API to support multi-query on the methods. The methods only accept arrays now.

## Usage

```ts
import algoliarecommend from '@algolia/recommend';

const recommendClient = algoliarecommend('appId', 'apiKey');

recommendClient
  .getRelatedProducts([
    {
      /* ... */
    },
  ])
  .then((response) => {
    console.log(response);
  });
```

### DIFF
--- a/packages/recommend/src/methods/getRecommendations.ts
+++ b/packages/recommend/src/methods/getRecommendations.ts
@@ -7,7 +7,7 @@ import {
   WithRecommendMethods,
 } from '../types';
 
-export type GetRecommendationsOptions = {
+export type GetRecommendationsQuery = {
   readonly indexName: string;
   readonly model: RecommendModel;
   readonly objectID: string;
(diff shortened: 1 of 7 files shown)


shortcuts (approved):
Looks clean!

packages/recommend/src/methods/getRecommendations.ts
```diff
@@ -22,16 +22,14 @@ type GetRecommendations = (
 ) => WithRecommendMethods<RecommendClient>['getRecommendations'];
 
 export const getRecommendations: GetRecommendations = base => {
```
shortcuts, line 24: Should we let users know that sending multiple queries with different model is not the expected behavior?

Haroenv (approved):
much better, glad you kept only one type (all plural)

packages/recommend/src/methods/getRecommendations.ts
francoischalifour: It would work as expected. However, in our AI components, we won't mix multiple models.

shortcuts: Ok!

PR #1284, opened 2021-06-23
fix(types): fix missing type exports

### LINKED ISSUE #1080: proposal: re-exports all types in `algoliasearch`

Haroenv:
Fixes #1080

This change marks all `@algolia/*` packages as bundledPackages for algoliasearch, which means that they are treated as part of this package and not external dependencies.

As a result API Extractor will throw an error if all types from `@algolia/*` packages that are accessible publicly through algoliasearch are not properly re-exported.

This PR is separated into 2 commits:

The 1st one updates API Extractor configuration to mark ae-forgotten-export messages as errors and updates the build script to mark all `@algolia/*` packages as bundledPackages (for algoliasearch only). This is done dynamically so we do not have to update the configuration if we add/remove some packages.
The 2nd one updates the actual TS files to fix the missing exports.


packages/algoliasearch/src/builds/node.ts
```diff
@@ -196,6 +196,148 @@ import { createUserAgent, RequestOptions } from '@algolia/transporter';
```
bodinsamuel, line 196: Can't we just do `export * from '@algolia/client-*'` ? 
Types listing seems unnecessary and hard to maintain + we would benefits from exporting helpers too

bodinsamuel (commented):
I thought it was done but I'm seeing some errors of missing export

packages/algoliasearch/src/builds/node.ts
bodinsamuel: (for `node.ts` only obviously)

Haroenv: seems like a good option, would you want to make a PR to take this over @bodinsamuel ? The only thing to be careful of with export * is that in some cases types are exported from JS files and then will have runtime errors when accessed directly from JS. I don't think that's really an issue here, but just to keep in mind

bobylito:
Hi there, what's the status? Is there something I can help with, maybe?

I use the latest version of the client and I can't access the type for the Settings and TS is bothering me with the `queryType` specifically.

shortcuts:
closing as this is done in v5

PR #1287, opened 2021-06-23
fix(recommend): export method types

francoischalifour:
This exports the method types for the Recommend API client.


shortcuts (approved):
Seems correct!

PR #1293, opened 2021-07-02
fix(answers): omit original hits in response

Haroenv:
Without this, the type is Hit[] & AnswersHit[] (paraphrased), which causes TypeScript to require extra casting to read _answer, which isn't expected.


alphonseb (approved):
LGTM

shortcuts (approved):

PR #1297, opened 2021-08-06
feat(auth): introduce WithinBody option for AuthMode

### LINKED ISSUE #1035: discussion: very long secured api-keys cause problems while searching
We use very long secured API keys, which works fine with the 3.x version of the package but version 4.x breaks this, as the key seems to be appended to the URL regardless how long it is. It basically reintroduces this: #319. I think version 4 should also sent long api keys via the POST body.

Haroenv:
fixes #1035

Implementation is done by:
- adding a new auth mode
- adding data as the return type for createAuth
- expose data from transporter
- serialize transporter data
- map api key back to query parameter if GET

This doesn't automatically switch to body if the key is too long, the option authMode needs to be set

```ts
algoliasearch('', '', { authMode: AuthMode.WithinBody })
```

TODO

- [ ] tests
- [ ] validate this works

### DIFF
--- a/packages/transporter/src/concerns/retryableRequest.ts
+++ b/packages/transporter/src/concerns/retryableRequest.ts
@@ -31,7 +31,7 @@ export function retryableRequest<TResponse>(
   /**
    * First we prepare the payload that do not depend from hosts.
    */
-  const data = serializeData(request, requestOptions);
+  const data = serializeData(transporter, request, requestOptions);
   const headers = serializeHeaders(transporter, requestOptions);
   const method = request.method;
 
@@ -40,6 +40,8 @@ export function retryableRequest<TResponse>(
     request.method !== MethodEnum.Get
       ? {}
       : {
+          // if AuthMode.WithinData, we forcibly map apiKey back to a query param
+          'x-algolia-api-key': transporter.data?.apiKey,
           ...request.data,
           ...requestOptions.data,
         };
--- a/packages/client-analytics/src/createAnalyticsClient.ts
+++ b/packages/client-analytics/src/createAnalyticsClient.ts
@@ -29,6 +29,8 @@ export const createAnalyticsClient: CreateClient<
       ...auth.queryParameters(),
       ...options.queryParameters,
     },
+
+    data: auth.data(),
   });
 
   const appId = options.appId;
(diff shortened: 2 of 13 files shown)


packages/transporter/src/concerns/retryableRequest.ts
```diff
@@ -40,6 +40,8 @@ export function retryableRequest<TResponse>(
     request.method !== MethodEnum.Get
       ? {}
       : {
+          // if AuthMode.WithinData, we forcibly map apiKey back to a query param
+          'x-algolia-api-key': transporter.data?.apiKey,
```
Haroenv, line 44: this might be implemented wrong?

tkrugg: Not sure what you mean by "wrong", but I have a extra question: is the idea here to move auth of all non GET requests to the body?

tkrugg (changes requested):
I agree with the change, but can we have a test securing this behaviour?

Haroenv:
it's in the todo @tkrugg

jpreynat:
@Haroenv Any news on this PR?
It'd be great to have this released to fix #1035

Haroenv:
Sorry, there was too many places to change in this unplanned PR, but withinHeaders should be an in-between solution, as it has a higher limit (but doesn't avoid a cors preflight request) @jpreynat

jpreynat:
Thanks for the suggestion @Haroenv. But sadly we already tried using `withinHeaders` and we have some cases where the limit is too small for us, preventing us from switching from version 3 to version 4.
Are there any plans on your side to have the `withinBody` option released at some point?

Haroenv:
At the moment it's not yet on the roadmap, as I made this pull request a while ago in an experiment. I will however ask the current owners of this repo whether they can take a look (@shortcuts, @millotp)

humanbagel:
@Haroenv @shortcuts @millotp @tkrugg can this prioritized now? it's really preventing enterprise users from the full performance benefits of algolia if we can't use long secured api keys from the browser

Haroenv:
Sorry, I'm not on this team anymore @humanbagel, but I do know that algoliasearch v3 had this feature, so you can use that before it's prioritised to be fixed. Someone who's actually on the api clients team can give more information on prioritisation

BohdanYavorskyi:
@nunomaduro what's the issue that this PR not merged? Enterprise-level apps need that a lot :(

Vishalbhilavala (approved):

Revadike:
Can we please merge this?

Revadike (approved):

PR #1300, opened 2021-08-25
fix(recommend): prevent `undefined` threshold

shortcuts:
**Summary**

Sending `threshold: undefined` to the API client causes the default value (`0`) to be overridden.

[sandbox reproduction](https://codesandbox.io/s/frosty-star-73kq7?file=/src/index.js)

**Result**

The default value isn't overridden when sending `threshold: undefined`

### DIFF
--- a/packages/recommend/src/__tests__/getRecommendations.test.ts
+++ b/packages/recommend/src/__tests__/getRecommendations.test.ts
@@ -126,7 +126,7 @@ describe('getRecommendations', () => {
     );
   });
 
-  test('override `undefined` threshold with default value', async () => {
+  test('overrides `undefined` threshold with default value', async () => {
     const client = createMockedClient();
 
     await client.getRecommendations(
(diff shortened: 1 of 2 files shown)


packages/recommend/src/__tests__/getRecommendations.test.ts
```diff
@@ -125,4 +125,40 @@ describe('getRecommendations', () => {
       {}
     );
   });
+
+  test('override `undefined` threshold with default value', async () => {
```
francoischalifour, line 129:
```suggestion
  test('overrides `undefined` threshold with default value', async () => {
```

francoischalifour (commented):
Can we add a test with a value for `threshold`?

packages/recommend/src/__tests__/getRecommendations.test.ts
shortcuts: fixed in https://github.com/algolia/algoliasearch-client-javascript/pull/1300/commits/3007cb4ca213e0f85fc87c3218ee1a9caf101768

francoischalifour (approved):
Thank you!

Haroenv:
What sets the threshold to undefined instead of nothing?

shortcuts:
> What sets the threshold to undefined instead of nothing?

Not sure if that's what you asked, but defaulted props [in the recommend UI library](https://github.com/algolia/recommend/blob/next/packages/recommend-core/src/getRecommendations.ts#L35-L43) for example, or if you send `threshold: undefined` to the client

PR #1305, opened 2021-09-13
docs(recommend): add TSDoc

francoischalifour:
This adds the TSDoc of the Recommend API client, based on the [existing documentation](https://www.algolia.com/doc/api-reference/api-methods/get-recommendations/).


Haroenv (approved):
👌 great contribution, thanks!

PR #1308, opened 2021-10-05
chore: dockerize development environment

millotp:
Ticket: [DI-14](https://algolia.atlassian.net/browse/DI-14)

In the sake of harmonisation and to ease development, this PR adds a docker file in which you can run the test or lint the code, while using your normal IDE.

They are a few caveats, the first one is that because we mount the project to the `/app` folder, the `node_modules` are shared and slow down the execution by a lot. To avoid that, the `node_modules` are installed in the parent folder (the root) and an empty volumes is mounted to avoid copying the pacakges from the host.  
This works fine but prevent the modification of the packages, and the image must be rebuilt when some are added.

I used [Docker Desktop](https://docs.docker.com/get-docker/) in the readme to install docker because docker machine is deprecated and the code is [archived](https://docs.docker.com/machine/).

To test the Dockerfile, follow the `DOCKER_README.MD` instructions and then run `yarn test:unit` for example.

### DIFF
--- a/DOCKER_README.MD
+++ b/DOCKER_README.MD
@@ -1,0 +1,51 @@
+In this page you will find our recommended way of installing Docker on your machine. 
+This guide is made for OSX users.
+
+## Install docker
+
+Install [Docker Desktop](https://docs.docker.com/get-docker/).
+
+## Build the image
+
+```bash
+docker build -t algolia-js .
+```
+
+## Run the image
+
+You need to provide few environment variables at runtime to be able to run the [Common Test Suite](https://github.com/algolia/algoliasearch-client-specs/tree/master/common-test-suite).
+You can set them up directly in the command:
+
+```bash
+docker run -it --rm --env ALGOLIA_APP_ID=XXXXXX [...] -v $PWD:/app -v /app/node_modules -w /app algolia-js bash
+```
+
+However, we advise you to export them in your `.bashrc` or `.zshrc`. That way, you can use [Docker's shorten syntax](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file) to set your variables.
+
+```bash
+### This is needed only to run the full test suite
+docker run -it --rm --env ALGOLIA_APPLICATION_ID_1 \
+                    --env ALGOLIA_ADMIN_KEY_1 \
+                    --env ALGOLIA_SEARCH_KEY_1 \
+                    --env ALGOLIA_APPLICATION_ID_2 \
+                    --env ALGOLIA_ADMIN_KEY_2
+                    --env ALGOLIA_APPLICATION_ID_MCM \
+                    --env ALGOLIA_ADMIN_KEY_MCM \
+-v $PWD:/app -v /app/node_modules -w /app algolia-js bash
+```
+
+Once your container is running, any changes you make in your IDE are directly reflected in the container, except for the `node_modules` folder.  
+If you want to add or remove packages, you will have to rebuild the image, this is because the `node_modules` folder is installed in a different folder to improve performance.
+
+To launch the tests, you can use one of the following commands
+```shell script
+# run only the unit tests
+yarn test:unit
+
+# run a single test
+yarn test:unit nameOfYourTest
+```
+
+You can find more commands in the `package.json` file.
+
+Feel free to contact us if you have any questions.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+# Dockerfile
+FROM node:10-alpine
+
+# Install the dependencies in the parent folder so they don't get overriden by the bind mount
+WORKDIR /
+
+# Needed to compile some npm packages
+RUN apk add --no-cache bash python3 make g++
+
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Dockerfile
+FROM node:10-alpine
+
+# Install the dependencies in the parent folder so they don't get overriden by the bind mount
+WORKDIR /
+
+# Needed to compile some npm packages
+RUN apk add --no-cache bash python3 make g++
```
Haroenv, line 8: which packages is this needed for? could we avoid them?

Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Dockerfile
+FROM node:10-alpine
```
Haroenv, line 2: should be harmonised with .nvmrc: https://github.com/algolia/algoliasearch-client-javascript/blob/master/.nvmrc

DOCKER_README.MD
```diff
@@ -0,0 +1,51 @@
+In this page you will find our recommended way of installing Docker on your machine. 
+This guide is made for OSX users.
```
Haroenv, line 2:
```suggestion
This guide is made for macOS users.
```

DOCKER_README.MD
```diff
@@ -0,0 +1,51 @@
+In this page you will find our recommended way of installing Docker on your machine.
```
Haroenv, line 1: this guide should be linked from the contributing file

Haroenv (commented):
👌

Dockerfile
millotp: `python3`, `make`, and `g++` are required for `iltorb`, a dependency of `bundlesize` and `rollup-plugin-filesize`, so it's dev only but the docker environment is meant for dev.

Haroenv: ah, could we maybe have a look at updating those versions of bundlesize & rollup-plugin-filesize? if I remember it correctly in their later versions they rely on the node builtin brotli support

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Dockerfile
-FROM node:10-alpine
+FROM node:12.16.2-alpine
```
Haroenv, line 2: can this be read from the nvmrc directly to not have a duplication?

millotp: You're right it's totally possible, once `renovate bot` will be installed it won't be an issue but it's a good idea for now.

Haroenv (dismissed):
good to go

Dockerfile
millotp: Looking deeper into it only `bundlesize` is requiring iltorb, but it doesn't seem to be maintained anymore, there is a [PR](https://github.com/siddharthkp/bundlesize/pull/370) opened for 6 months, and another [package](https://github.com/siddharthkp/bundlesize2) that solves that and doesn't require any `node-gyp`.

Haroenv: let's leave it as-is then for now @millotp :)

Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Dockerfile
+ARG NODE_IMAGE=node:12.16.0-alpine
+
+FROM $NODE_IMAGE
+
+# Install the dependencies in the parent folder so they don't get overriden by the bind mount
+WORKDIR /
+
+# Needed to compile some npm packages
```
DevinCodes, line 9:
```suggestion
# We need to install some dependencies for bundlesize (https://github.com/siddharthkp/bundlesize/pull/370)
```

DOCKER_README.MD
```diff
@@ -0,0 +1,51 @@
...
+
+However, we advise you to export them in your `.bashrc` or `.zshrc`. That way, you can use [Docker's shorten syntax](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file) to set your variables.
+
+```bash
+### This is needed only to run the full test suite
+docker run -it --rm --env ALGOLIA_APPLICATION_ID_1 \
+                    --env ALGOLIA_ADMIN_KEY_1 \
+                    --env ALGOLIA_SEARCH_KEY_1 \
+                    --env ALGOLIA_APPLICATION_ID_2 \
+                    --env ALGOLIA_ADMIN_KEY_2
```
DevinCodes, line 31:
```suggestion
                    --env ALGOLIA_ADMIN_KEY_2 \
```

DevinCodes (changes requested):
This is great, thank you @millotp. It's almost good to go! 🚀

millotp:
Thanks @Haroenv and @DevinCodes for the review !

Haroenv:
The test is failing because the node docker container of circleCI that doesn't have the latest node versions is bundled with an outdated ssl implementation that resolves lets encrypt wrong, and doesn't allow yarn installs

DevinCodes:
@Haroenv is there anything that we can do to resolve that? Or do we depend on CircleCI to resolve this in their image?

Haroenv:
What do you think of harmonising docker image in CI and this docker image in some later step?

PR #1312, opened 2021-10-15
feat: add custom request function APIC-126

millotp:
Jira ticket: [APIC-126](https://algolia.atlassian.net/browse/APIC-126)

Adds the `customRequest` function to the search client for all environment, to allows clients to use features that are not yet implemented.

I didn't find a way to automatically generate the `algoliasearch/src/builds/*.ts` files, I'm not sure if there is one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     },
     {
       "path": "packages/algoliasearch/dist/algoliasearch-lite.umd.js",
-      "maxSize": "4.35KB"
+      "maxSize": "4.4KB"
     },
     {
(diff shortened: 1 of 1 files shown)


millotp:
It's done in Python and for Android, and it's called `custom_request`. This is part of the [DI sprint](https://algolia.atlassian.net/jira/software/c/projects/DI/boards/171)

Haroenv:
Makes sense. In theory js could still be an outlier, as it usually favours shorter method names, but I'll leave it up to you

PR #1314, opened 2021-10-20
ci(circle): use nvm for node unit tests

Haroenv:
the node 8 image was outdated and causes certificate issues


shortcuts:
oki I lost :(

PR #1370, opened 2021-11-10
chore: disable renovate until the test suite is less flaky

millotp:
Disable renovate until the CTS is in place, or some other way to make the test less flaky and able to run in parallel.


bodinsamuel (approved):
fair enough

PR #1371, opened 2021-11-10
feat(node): dedupe in-flight network requests by default

Haroenv:
I'm wondering if we shouldn't even enable the responses cache as well, but at least the requestsCache is an obvious choice

Could be considered a breaking change


millotp (approved):
What's the motivation to enable the cache ?
I guess it's fine because it's already setup in the others environment, but the cache should have a lifespan

Haroenv:
hmm yes, this could introduce failures in high-throughput scenarios, I'll leave this as-is for now, but if you feel more comfortable or in a major version, you can pick this back up

millotp:
Thanks we will think about it :)

Issue #1383, opened 2022-01-24
SearchOptions.facetFilters type is too restrictive


eliasnorrby:
As per the [docs][1], `SearchOptions.facetFilters` can be a list with elements being a list of `string | string[]`:

```js
index.search('', {
  facetFilters: [["author:Stephen King", "genre:Horror"], "publisher:Penguin"]
});
```

The [currently exported type][2], however, requires all elements to be of the same type, i.e. either `string` or `string[]`:

```ts
  readonly facetFilters?: string | readonly string[] | ReadonlyArray<readonly string[]>;
```

Attempting to call `client.search` with an array of mixed contents results in this error:

```ts
 error TS2322: Type '(string | string[])[]' is not assignable to type 'string | readonly string[] | readonly (readonly string[])[]'.
  Type '(string | string[])[]' is not assignable to type 'readonly (readonly string[])[]'.
    Type 'string | string[]' is not assignable to type 'readonly string[]'.
      Type 'string' is not assignable to type 'readonly string[]'.

           { facetFilters: ['abc:123', ['def:456','one:two']] },
```

[1]: https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/in-depth/filters-and-facetfilters/
[2]: https://github.com/algolia/algoliasearch-client-javascript/blob/2c626a3dd59941b97549990bc03af4dd9b095484/packages/client-search/src/types/SearchOptions.ts#L17

PR #1384, opened 2022-01-24
fix(types): allow mix of strings and arrays in facetFilters

eliasnorrby:
fixes: #1383


Haroenv (approved):
great catch, thanks!

bs-twic:
Hey! Any idea when this will be released to NPM?

PR #1394, opened 2022-02-22
# fix(types): updating reRankingApplyFilter type and adding new property to RankingInfo

alphonseb:
# fix(types): updating reRankingApplyFilter type and adding new property to RankingInfo

This is a follow up to #1378 

## What
1. Adding `null` to the list of possible types for the `reRankingApplyFilter` search/setting option
1. Adding property `promotedByReRanking` to the `RankingInfo` type.

## Why
1. This is how we reset the filter in the dashboard, it was an oversight in the first PR.
2. Has been sent back by the engine for some time but was never added to the client's types.

### DIFF
--- a/packages/client-search/src/types/Hit.ts
+++ b/packages/client-search/src/types/Hit.ts
@@ -43,6 +43,7 @@ export type RankingInfo = {
     readonly rankingScore: number;
     readonly score: number;
   };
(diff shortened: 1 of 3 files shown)


Haroenv (approved):
Has this been added to the api client spec?

PR #1395, opened 2022-02-24
fix(types): allow mixed arrays in numeric and tag filters

eliasnorrby:
Correct me if I'm wrong, but I believe the fix for `facetFilters` in #1384 should apply to `numericFilters` and `tagFilters` as well.


Haroenv (approved):
thanks, that's correct!

eliasnorrby:
@Haroenv As with #1384, given that this repo does not have automatic releases, would you be able to trigger one manually? 😇

shortcuts:
Hey @eliasnorrby, it will be released once https://github.com/algolia/algoliasearch-client-javascript/pull/1397 is merged

PR #1406, opened 2022-05-17
feat(search): add type for queryLanguages

### LINKED ISSUE #1405: queryLanguages not listed in SearchOptions
According to the [documentation](https://www.algolia.com/doc/api-reference/api-parameters/queryLanguages/) of the `queryLanguages` parameter, it can be passed to the search function as so:

```
index.search('query string', { queryLanguages: ['en'] })
  .then(({ hits }) => {
    console.log(hits);
  });
```

The second parameter of `search` is `requestOptions` and is typed as `RequestOptions & SearchOptions`. However I don't see the `queryLanguages` listed in the `SearchOptions` type (I don't expect it to be in `RequestOptions` since that type is purely for configuring the request itself). Right now `queryLanguages` is listed in the in `Settings` type but I believe it could also be added to `SearchOptions`. 

What do you guys think about adding `queryLanguages` to `SeachOptions`?

Haroenv:
According to https://www.algolia.com/doc/api-reference/api-parameters/queryLanguages/?client=javascript, it can be sent as a search parameter too.

fixes #1405

### DIFF
--- a/packages/client-search/src/types/SearchOptions.ts
+++ b/packages/client-search/src/types/SearchOptions.ts
@@ -368,4 +368,9 @@ export type SearchOptions = {
     | readonly string[]
     | ReadonlyArray<readonly string[] | string>
     | null;
+
+  /**
+   * Sets the languages to be used by language-specific settings and functionalities such as ignorePlurals, removeStopWords, and CJK word-detection.
+   */
+  readonly queryLanguages?: readonly string[];
 };


shortcuts (approved):
Good catch!

Haroenv:
the test failures are unrelated, @shortcuts, will you follow up before release? they're required to pass before it can be released :)

Issue #1418, opened 2022-07-20
Missing property "abTestID" in search response TypeScript types


anjakunkel:
I would like to report a missing field in your TypeScript type definitions:
According to https://www.algolia.com/doc/api-reference/api-parameters/getRankingInfo/?client=javascript, I expect a field **abTestID** to be part of the **SearchResponse** type.

When looking into the JSON response of the request, the "abTestID" field is present and contains correct data. The only part missing is that is not in your TypeScript type definitions, yet.

PR #1419, opened 2022-07-27
fix(types): add `abTestID` to `SearchResponse`

shortcuts:
## Summary

Ticket: https://algolia.atlassian.net/browse/CR-1581

Fixes https://github.com/algolia/algoliasearch-client-javascript/issues/1418

Add `abTestID` field to the `SearchResponse` type.

### DIFF
--- a/packages/client-search/src/types/SearchResponse.ts
+++ b/packages/client-search/src/types/SearchResponse.ts
@@ -156,6 +156,11 @@ export type SearchResponse<TObject = {}> = {
   indexUsed?: string;
 
   /**
+   * If a search encounters an index that is being A/B tested, abTestID reports the ongoing A/B test ID.
+   */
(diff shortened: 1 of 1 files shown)


shortcuts:
Looking at https://www.algolia.com/doc/api-reference/api-parameters/getRankingInfo/#impact-on-the-response it's a bit confusing if it's under `_rankingInfo` or at the root, we've put it at the root in v5 which makes sense as it's bound to `abTestVariantID`, wdyt?

bodinsamuel (approved):
Testing gives me _rankingInfo in the hits and other fields at the root. The doc is indeed not perfectly clear.
Screenshot 2022-07-27 at 09 46 29

shortcuts:
Yep indeed, I'll add a ticket in our maintenance backlog

PR #1433, opened 2022-12-15
fix(recommend): remove unused trendingFacets props 

marialungu:
Remove `fallbackParameters` and `queryParameters` given that the recommend API does not support these props.
🐛 : [JIRA](https://algolia.atlassian.net/browse/RECO-1173)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,3 @@ dist
 *.umd.js
 yarn-error.log
 .env
-
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ dist
 *.umd.js
 yarn-error.log
 .env
+
+.idea/
--- a/packages/recommend/src/types/TrendingFacetsQuery.ts
+++ b/packages/recommend/src/types/TrendingFacetsQuery.ts
@@ -1,3 +1,3 @@
 import { TrendingQuery } from './TrendingQuery';
 
-export type TrendingFacetsQuery = Omit<TrendingQuery, 'model' | 'facetValue'>;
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -6,3 +6,5 @@ dist
 *.umd.js
 yarn-error.log
 .env
+
+.idea/
```
jojva, line 10: Make a separate PR for this

Haroenv: it probably should just be reverted actually, local files should be in a global gitignore

Haroenv (commented):
change itself looks legit

packages/recommend/src/types/TrendingFacetsQuery.ts
```diff
@@ -1,3 +1,6 @@
 import { TrendingQuery } from './TrendingQuery';
 
-export type TrendingFacetsQuery = Omit<TrendingQuery, 'model' | 'facetValue'>;
+export type TrendingFacetsQuery = Omit<
+  TrendingQuery,
+  'model' | 'facetValue' | 'fallbackParameters' | 'queryParameters'
```
francoischalifour, line 5: At this point why have a shared `TrendingQuery` type if they differ so much?

According to the API endpoints:

- Trending facets
  - `facetName`: required
- Trending items
  - `facetName`: optional
  - `facetValue`: optional

jojva: Yeah that's a good point

marialungu: yeah, i wanted to do that in a separate PR, both for client and for the libs

jojva (approved):

PR #1434, opened 2022-12-16
fix(recommend): update trending facets query props (RECO-1184)

marialungu:
Update trendingFacets query props and refactor types


packages/recommend/src/types/TrendingFacetsQuery.ts
```diff
@@ -1,6 +1,21 @@
...
+   */
+  readonly threshold?: number;
+
+  /**
+   * How many recommendations to retrieve.
+   */
+  readonly maxRecommendations?: number;
+
+  /**
+   * Used for trending model
```
francoischalifour, line 18: This doesn't describe what this param is? (Same for `facetName` and `facetValue` for trending items.)

packages/recommend/src/types/index.ts
```diff
@@ -11,6 +11,5 @@ export * from './RecommendSearchOptions';
 export * from './RecommendationsQuery';
 export * from './RelatedProductsQuery';
 export * from './TrendingFacetsQuery';
-export * from './TrendingQuery';
```
francoischalifour, line 14: Any reasons to remove this export?

marialungu: removed the file completely and separated the trending query in 2 independent ones

francoischalifour: Yes but the type still exists: https://github.com/algolia/algoliasearch-client-javascript/pull/1434/files#diff-c2f0ecb57e8ccf256bade6bed628a6494a8758ed3a4756877bcccf1b3db25d69R16-R18

I'm fine removing the export as long as it's a conscious decision.

marialungu: but here is only a local type as it's only used (and needed) in `getRecommendations`, so there's no need to export it.

francoischalifour: I was more hinting the fact that it could be a breaking change now that it's not exported anymore. I think it's OK, but we should acknowledge it.

francoischalifour (approved):

PR #1438, opened 2023-02-01
feat(ts): add redirect to rendering content

drodriguln:
Adds types to support the new supported redirect payload within `renderingContent`. [See the RFC](https://algolia.atlassian.net/wiki/spaces/FX/pages/4459495428/UI+Platform+-+RFC+OOTB+Redirects) for more details.

### DIFF
--- a/packages/client-search/src/types/Settings.ts
+++ b/packages/client-search/src/types/Settings.ts
@@ -342,9 +342,6 @@ export type Settings = {
         };
       };
     };
-    /**
-     * defining out-of-the-box redirect
-     */
     readonly redirect?: {
       /**
        * the url for the client to redirect to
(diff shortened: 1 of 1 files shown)


packages/client-search/src/types/Settings.ts
```diff
@@ -342,6 +342,15 @@ export type Settings = {
         };
       };
     };
+    /**
+     * defining out-of-the-box redirect
+     */
```
Haroenv, line 347:
```suggestion
```

drodriguln:
Doesn't look like this is necessary now for the time being. Closing this PR.

Haroenv:
It's not necessary for the plugin indeed, but we will want to add it later anyway, right?

PR #1443, opened 2023-03-07
Fix for the redirect object in SearchResponse

sfaiqh:


Haroenv (approved):
this seems correct, is it live?

packages/client-search/src/types/SearchResponse.ts
```diff
@@ -224,6 +225,10 @@ export type SearchResponse<TObject = {}> = {
      * as a JSON object with one field per parameter.
      */
     params?: Record<string, any>;
+
+    redirect?: {
```
shortcuts, line 229: if we have a link to the public doc, it would make sense to me to add it here

shortcuts (commented):
same question as @Haroenv!

packages/client-search/src/types/SearchResponse.ts
sfaiqh: This parameter is not public and is not communicated to customers.

sfaiqh:
> 



> this seems correct, is it live?

Yes

sfaiqh:
@Haroenv anything else needed on this PR?

packages/client-search/src/types/SearchResponse.ts
sfaiqh: @shortcuts added a comment for internal use

shortcuts (approved):
looks good! :)

PR #1452, opened 2023-03-23
FEAT: Removing Mode for CloudFlare Workers

OLingard:
## Description

When trying to use Algolia's `requester-fetch` package combined with the `algoliasearch` method on Cloudflare workers, we get the error `The 'mode' field on 'RequestInitializerDict' is not implemented.`. For reference CF workers are ran in V8 Isolate and they have not added an implementation for mode on their fetch runtime.

Although it's not ideal that the behaviour of CloudFlare is to throw an error, we can avoid this issue by removing the mode field from the `requester-fetch` fetch options. The default mode is `cors` anyway so there is no functional change.

This issue has surfaced for multiple people as seen in [this forum](https://discourse.algolia.com/t/cloudflare-workers-and-algoliasearch-javascript-problem/16714).

## Reasons

I looked at the other request packages but they do not fit the bill.

- `requester-node-http`: CF workers are ran in a browser environment not node
- `requester-browser-xhr`: `XMLHttpRequest` is not implemented on CF workers.

### DIFF
--- a/packages/requester-fetch/src/createFetchRequester.ts
+++ b/packages/requester-fetch/src/createFetchRequester.ts
@@ -44,7 +44,6 @@ export function createFetchRequester({
             ...request.headers,
           },
           body: request.data || null,
-          mode: 'cors',
           redirect: 'manual',
           signal,
         });


packages/requester-fetch/src/createFetchRequester.ts
shortcuts, line 1: hey, instead of removing the feature (which would be a breaking change), you can change the order of the parameters, like we did in v5 https://github.com/algolia/api-clients-automation/blob/main/clients/algoliasearch-client-javascript/packages/requester-fetch/src/createFetchRequester.ts#L39

so that you can then override the value

shortcuts (commented):
thanks for the PR!

packages/requester-fetch/src/createFetchRequester.ts
OLingard: Hey @shortcuts, I'm not getting why this is a breaking change. The functionality has not changed in the slightest. As shown [here](https://javascript.info/fetch-api), the default mode is 'cors' so shouldn't we be able to remove this field with no effect? The order of the parameters doesn't help me I think as the issue is that the field mode exists as an option, not where it is placed in the object.

shortcuts: > I'm not getting why this is a breaking change. The functionality has not changed in the slightest. As shown [here](https://javascript.info/fetch-api), the default mode is 'cors' so shouldn't we be able to remove this field with no effect?

but if this comes to change at some point, we will not control it so could be an unwanted behavior. We are already planning breaking changes in v5, so I'd personally be fine by removing it in the next major, but would leave it as is for now

> The order of the parameters doesn't help me I think as the issue is that the field mode exists as an option, not where it is placed in the object.

changing the order and putting the spread of the requestOptions (like here https://github.com/algolia/api-clients-automation/blob/main/clients/algoliasearch-client-javascript/packages/requester-fetch/src/createFetchRequester.ts#L45) allows you to provide a `mode: null` or `mode: undefined` value to override it

OLingard: > but if this comes to change at some point, we will not control it so could be an unwanted behavior. We are already planning breaking changes in v5, so I'd personally be fine by removing it in the next major, but would leave it as is for now

Just on this, I'm not sure if this is a warranted concern. We already have this issue for many other options inside the fetch request. If we look at a list of fetch options with their defaults:
```js
referrer: "about:client", // or "" to send no Referer header,
// or an url from the current origin
referrerPolicy: "strict-origin-when-cross-origin", // no-referrer-when-downgrade, no-referrer, origin, same-origin...
mode: "cors", // same-origin, no-cors
credentials: "same-origin", // omit, include
cache: "default", // no-store, reload, no-cache, force-cache, or only-if-cached
integrity: "", // a hash, like "sha256-abcdef1234567890"
keepalive: false, // true
signal: undefined, // AbortController to abort request
window: window // null
```
We already have this worry for cache, credentials, keepalive, referrerPolicy etc.

Given that fetch is so widely used and this change is very unlikely given the ramifications for almost every package which requests data, are we okay to remove it seeing that multiple people have reported this as an issue?

OLingard: If their is still worry, I can make the change to adjust the requester options placement, I just am trying to help other people so they don't have to go through the same debugging process

OLingard: Hi @shortcuts, just bumping this to get your further thoughts

alexkraj: Hi @shortcuts , just wanted to check-in again on the above

OLingard: Thanks @shortcuts! I'm not sure of the contributing process, I think everything is complete from my side but just want to confirm that's the case?

shortcuts (approved):
ran everything locally and it works! thanks :D

mglikesbikes:
hello! wanted to ask a clarifying question: does this PR enable Cloudflare Workers 100%, no other workarounds etc?

Separately, noticed there's a new beta version of the JS library, will that include this fix _or_ has the beta no necessarily been tested on Workers platform yet?

Thanks in advance!

shortcuts:
Hey!

> hello! wanted to ask a clarifying question: does this PR enable Cloudflare Workers 100%, no other workarounds etc?

I believe so, but if not feel free to submit an issue and we would be more than happy to resolve it :)

> Separately, noticed there's a new beta version of the JS library, will that include this fix _or_ has the beta not necessarily been tested on Workers platform yet?

It have already been removed from the new beta version (v5), and the requester is exactly at the same stage as this one, however I don't know if people already tested it on workers

mglikesbikes:
thanks so much!

PR #1453, opened 2023-03-24
feat(neural): make eventSources nullable

cdhawke:
## Why

Explicit nullable type for the `eventSources` setting. This is helpful since the `null` state is the 'default' for the field, and nulling the value has specific behavior in the search engine (e.g. to use the primary/replica index group as event sources).


febeck (approved):

shortcuts (approved):
looks great :) it would be great to add some tests for those new parameters

PR #1460, opened 2023-05-16
fix(types): union SearchForFacetValuesResponse type for multipleQueries

omikader:
## Summary

This PR unions `SearchForFacetValuesResponse` with `SearchResponse<TObject>` as possible return types for results in the `multipleQueries` API.

## Context

The `MultipleQueriesQuery` interface [states](https://github.com/algolia/algoliasearch-client-javascript/blob/4.13.0/packages/client-search/src/types/MultipleQueriesQuery.ts#L27-L48) that you can supply either a `default` or a `facet` query. This is also confirmed by the Algolia [documentation.](https://www.algolia.com/doc/api-reference/api-methods/multiple-queries/#method-param-type)

However, the MultipleQueriesResponse [definition](https://github.com/algolia/algoliasearch-client-javascript/blob/4.13.0/packages/client-search/src/types/MultipleQueriesResponse.ts#LL7C18-L7C32) does not include [SearchForFacetValuesResponse](https://github.com/algolia/algoliasearch-client-javascript/blob/4.13.0/packages/client-search/src/types/SearchForFacetValuesResponse.ts) as a possible return type. This is confusing because the response data from Algolia does in fact include facet responses when requested (see attached image).

image

### DIFF
--- a/packages/client-search/src/__tests__/integration/multiple-operations.test.ts
+++ b/packages/client-search/src/__tests__/integration/multiple-operations.test.ts
@@ -1,6 +1,12 @@
 import { BatchActionEnum, StrategyEnum } from '../..';
 import { TestSuite } from '../../../../client-common/src/__tests__/TestSuite';
-import { MultipleBatchRequest } from '../../types';
+import { MultipleBatchRequest, MultipleQueriesResponse, SearchResponse } from '../../types';
+
+function expectSearchResponse<TObject>(
+  results: MultipleQueriesResponse<TObject>['results'][number]
+): asserts results is SearchResponse<TObject> {
+  expect(results).toHaveProperty('hits');
+}
 
 const testSuite = new TestSuite('multiple_operations');
 
@@ -62,7 +68,9 @@ test(testSuite.testName, async () => {
     { strategy: StrategyEnum.None }
   );
 
+  expectSearchResponse(multipleQueriesResponse1.results[0]);
   expect(multipleQueriesResponse1.results[0].hits).toHaveLength(2);
+  expectSearchResponse(multipleQueriesResponse1.results[1]);
   expect(multipleQueriesResponse1.results[1].hits).toHaveLength(2);
 
   const multipleQueriesResponse2 = await client.search(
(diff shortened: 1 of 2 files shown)


shortcuts:
@Haroenv do you know why we did not made the union in the first place?

Haroenv (approved):
While I feared this was a breaking change, as only `MultipleQueriesResponse` is changed, this actually does not impact any code used by customers today (only the helper's internals, but that isn't typescript). I'm thus happy with the change after all. Sorry for the delay!

shortcuts:
> While I feared this was a breaking change, as only `MultipleQueriesResponse` is changed, this actually does not impact any code used by customers today (only the helper's internals, but that isn't typescript). I'm thus happy with the change after all. Sorry for the delay!

good to know! thanks for double checking

chrisneven:
@Haroenv fyi, it did impact us in a way:

image

We depend on this typing a lot throughout our codebase

anjakunkel:
Indeed, the type change has breaking impact, also on us.

But it can be fixed easily by explicitly typing your `searchResult` to be either `SearchForFacetValuesResponse` with `SearchResponse<TObject>`, depending on your input.

VictorGosse:
Hello,

We are indeed facing the same typing issue after the upgrade to `4.18.0`. As suggested above we can explicitly type the `searchResult` (I guess via a `as` ?).

Couldn't it be possible to improve the typing here by allowing the type to be sent as a generic so that way it could be defined directly when calling `multipleQueries` and would be type safe.
eg: `client.multipleQueries<SearchResponse<TObject>>(...)` would force the return type to be `SearchResponse<TObject>`.
As there is already a generic on `multipleQueries` maybe it could be enough to use it: if `TObject` is defined then it'll return `SearchResponse<TObject>` otherwise it could be both.

Or even better maybe the return type could be inferred based on the data sent in `multipleQueries` 🤷‍♂️ 

Thanks :)

ranman:
This also broke our search.

How can we import the relevant types from the `algoliasearch` namespace or do we need to add the @algolia/* packages into the imports?

```typescript
export const Algolia = {
  async search<T = Record<string, string>[]>(
    query: string,
    params: Record<string, string | number> = {}
  ): Promise<{ hits: T; totalPages: number }> {
    try {
      const queries = BUILDER_MODELS.map((model) => {
        return {
          indexName: model,
          query,
          params: {
            ...params,
            hitsPerPage: HITS_PER_PAGE,
          },
        }
      })

      const res = await client.multipleQueries(queries)

      // Retrieve hits out of results and return
      // @ts-expect-error
      const hits = res?.results?.flatMap((result) => result.hits)
      // @ts-expect-error
      const pages = Math.max(...res.results.map((result) => result.nbPages))
      return { hits: hits as T, totalPages: pages - 1 }
    } catch (e) {
      throw new Error(`Failed to load search data`)
    }
  },
}
```

BohdanYavorskyi:
This is definitely the breaking change and MUST be updated because `as Type` does not work without `as unknown as`

PR #1469, opened 2023-07-10
feat(recommend): add "Looking Similar" model

raed667:
Implement Recommend `looking-similar` model. 

Beyond the model name, everything else should work exactly as other recommend models such as related products.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "4.18.0"
+  "version": "4.18.1-rc.1"
 }
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "4.18.1-rc.1"
+  "version": "4.18.0"
 }
--- a/packages/recommend/src/methods/getLookingSimilar.ts
+++ b/packages/recommend/src/methods/getLookingSimilar.ts
@@ -3,7 +3,7 @@ import { getRecommendations } from './getRecommendations';
 
 type GetLookingSimilar = (
   base: BaseRecommendClient
-) => WithRecommendMethods<BaseRecommendClient>['getRelatedProducts'];
+) => WithRecommendMethods<BaseRecommendClient>['getLookingSimilar'];
 
 export const getLookingSimilar: GetLookingSimilar = base => {
   return (queries: readonly LookingSimilarQuery[], requestOptions) => {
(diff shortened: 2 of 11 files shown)


packages/recommend/src/methods/getLookingSimilar.ts
```diff
@@ -0,0 +1,18 @@
+import { BaseRecommendClient, LookingSimilarQuery, WithRecommendMethods } from '../types';
+import { getRecommendations } from './getRecommendations';
+
+type GetLookingSimilar = (
+  base: BaseRecommendClient
+) => WithRecommendMethods<BaseRecommendClient>['getRelatedProducts'];
```
dhayab, line 6: Might not be related to the failing tests, but should this be like this instead?

```suggestion
) => WithRecommendMethods<BaseRecommendClient>['getLookingSimilar'];
```

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "4.18.0"
+  "version": "4.18.1-rc.1"
```
dhayab, line 4: Version update is managed by Ship.js so you don't need to manually set your new version (same for `packages/client-common/src/version.ts` and `packages/recommend/package.json`).

I'm not sure of the way to go if you want to release a RC version of the recommend client. One simple way is to stay in this feature branch and publish manually. This way you'll only release `@algolia/recommend@4.18.1-rc.1` and not all of the other packages in the same monorepo. Probably need a validation from the API clients owners.

dhayab (commented):
My guess regarding failing tests is that they are just really flaky. I looked into merged PRs and all their pipelines are retried a few time until they go green.

lerna.json
raed667: Yes apologies, this was not meant to be for the actual PR. 

It was just to check if my CodeSandbox was pulling the right version. I'll revert the change before setting the PR as open.

packages/recommend/src/methods/getLookingSimilar.ts
raed667: Good catch!

dhayab (approved):
Looks good to me 👍.

Regarding the RC release, as this addition follows closely how other methods are defined, I think it can be released directly in a minor as part of the monorepo release process.

If you prefer having an RC/Beta release, you should not merge this branch, and we can instead perform a manual release (or perhaps there's another process owners can guide us through).

raed667:
@shortcuts indeed we should this to the generated clients as well 👍 

@dhayab No beta/RC needed, once merged this should probably be simply released under `4.19.0`

shortcuts:
> @shortcuts indeed we should this to the generated clients as well 👍

cool! let me know if you need help contributing to the specs

marialungu (approved):

PR #1472, opened 2023-07-18
fix(recommend): Add Type RecommendQueriesResponse from old MultipleQueriesResponse for recommend

raed667:
In recommend add type `MultipleQueriesResponse` that basically reverts back to the old definition before PR https://github.com/algolia/algoliasearch-client-javascript/pull/1460

### DIFF
--- a/packages/recommend/src/types/WithRecommendMethods.ts
+++ b/packages/recommend/src/types/WithRecommendMethods.ts
@@ -9,10 +9,7 @@ import { TrendingFacetsQuery } from './TrendingFacetsQuery';
 import { TrendingItemsQuery } from './TrendingItemsQuery';
 import { TrendingQuery } from './TrendingQuery';
 
-/**
- * This MultipleQueriesResponse type differs from MultipleQueriesResponse imported from @algolia/client-search as it omits "SearchForFacetValuesResponse" in results
- */
-export type MultipleQueriesResponse<TObject> = {
+export type RecommendQueriesResponse<TObject> = {
   /**
    * The list of results.
    */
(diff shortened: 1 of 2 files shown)


packages/recommend/src/types/WithRecommendMethods.ts
```diff
@@ -9,6 +9,16 @@ import { TrendingFacetsQuery } from './TrendingFacetsQuery';
 import { TrendingItemsQuery } from './TrendingItemsQuery';
 import { TrendingQuery } from './TrendingQuery';
 
+/**
+ * This MultipleQueriesResponse type differs from MultipleQueriesResponse imported from @algolia/client-search as it omits "SearchForFacetValuesResponse" in results
+ */
+export type MultipleQueriesResponse<TObject> = {
```
Haroenv, line 15: this should have a different name and probably a test to ensure it's caught by ci

raed667: Renamed to `RecommendQueriesResponse`

Haroenv (dismissed):

Haroenv (commented):
however, this should have a test to avoid the issue you make the PR for

raed667:
@Haroenv added a test that fails when the old type is used but passes when the new one is set.

Haroenv (approved):

shortcuts (approved):
gg!

packages/recommend/src/types/WithRecommendMethods.ts
lyczkop: this RecommendQueriesResponse shouldn't be used also in javascript client: `packages/client-search/src/methods/client/multipleQueries.ts` ?
because #1460 had impact also on this method

PR #1490, opened 2023-11-23
feat(recommend): add recommended-for-you model

raed667:
## Description 

Add `getRecommendedForYou` function to recommend.

## Test

To test locally you can update `playground/browser/index.html` with the following content: 

```html 
<html>
  <head>
    <script src="/playground/browser/recommend.umd.js"></script>

    <script>
      const recommend = window['@algolia/recommend'];
      const client = recommend('betaUASF0EMRRM', '2c80241f06757af6fc4ef4a2d9523b6d');

      client
        .getRecommendedForYou([
          {
            indexName: 'bestbuy',
            queryParameters: {
              userToken: 'user_token_1',
            },
          },
        ])
        .then(({ results }) => {
          console.log(results[0]);
        })
        .catch(err => {
          console.error(err);
        });
    </script>
  </head>
  <body></body>
</html>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     },
     {
       "path": "packages/recommend/dist/recommend.umd.js",
-      "maxSize": "4.32KB"
+      "maxSize": "4.4KB"
     }
   ]
 }
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     },
     {
       "path": "packages/recommend/dist/recommend.umd.js",
-      "maxSize": "4.3KB"
+      "maxSize": "4.32KB"
     }
   ]
 }
--- a/packages/client-common/src/__tests__/TestSuite.ts
+++ b/packages/client-common/src/__tests__/TestSuite.ts
@@ -39,7 +39,7 @@ export class TestSuite {
   public indicesCount = 0;
 
   public constructor(testName?: string) {
-    // this.ensureEnvironmentVariables();
+    this.ensureEnvironmentVariables();
 
     this.testName = testName || '';
   }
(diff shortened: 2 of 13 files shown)


marialungu (dismissed):
LGTM 🚀

francoischalifour:
There are 3 ways of referencing "products" now:

- Related **Products**
- Trending **Items**
- Recommended For You (generic, no name)

What's the reasoning behind all these different namings? "Recommended For You" also returns products, right? And in that case, how to explain users the difference between "Recommended For You" and "Related Items" provided with a `userToken`?

AlexEven:
Hey François, will try my best to answer your questions

* **Related products**: products that are linked to the source item in a way (either it's an alternative, or it's often be sold together, etc)
* **Trending items**: product in your catalog that have recently become popular
* **Recommended for you**: a subset of the catalog matching your affinities. E.g. on homepage, pick most relevant items from the full catalog; on a collection page, highlight personalized items from a specific category; on user profile page, show personalized items from best sellers, whatever. Possibilities are endless.

**How is this different from Related products with a user token?**
This second option is "simply" re-ranking already available recommendations to push the ones matching user affinities in priority. Just like you would apply a rule to boost a certain category, here you'd prioritize personalization.

Hope this helps. Thanks!

francoischalifour:
Right, so the main difference with Recommended For You is that there's no reference product. Thanks for the clarification!

(It's unfortunate that there's this divergence between Related _Products_ and Trending _Items_, while both refer to catalog records.)

francoischalifour (approved):

marialungu (approved):
🚀

raed667:
@shortcuts this can be merged

PR #1494, opened 2023-12-05
Fix/recommend get trending facets typing

marialungu:
The `hits` returned by `getTrendingFacets` differ (they only have `faceValue`, `facetName` and `_score`) from the normal hits returned by the other recommendation methods.


raed667 (approved):
🥇 LGTM 🥇

jojva (approved):
LGTM :+1:

PR #1498, opened 2023-12-14
fix(recommend): update TrendingFacetHit facetValue type to string

raed667:
The `trending-facets` API will only return type `string` for `facetValue`. 

To simplify the code, we can remove the type argument `TObject` for `trending-facets`

### DIFF
--- a/packages/recommend/src/types/TrendingFacetHit.ts
+++ b/packages/recommend/src/types/TrendingFacetHit.ts
@@ -1,5 +1,5 @@
-export type TrendingFacetHit<TObject> = {
+export type TrendingFacetHit = {
   readonly _score: number;
   readonly facetName: string;
-  readonly facetValue: TObject;
+  readonly facetValue: string;
 };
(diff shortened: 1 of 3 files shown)


packages/recommend/src/types/TrendingFacetHit.ts
```diff
@@ -1,5 +1,5 @@
-export type TrendingFacetHit<TObject> = {
+export type TrendingFacetHit = {
```
Haroenv, line 1: if this type is exported, it's a breaking change, someone who would be using `TrendingFacetHit<string>` will get a typescript error. Instead all generics should stay, defaulted to `string`

raed667: IMO it is fine to have a build step breaking change in this case as it is a bug fix in typings.

Plus the usage for this model is very small and the update will be a one liner documented in the release notes.

Haroenv (approved):
BC ok because of low impact

shortcuts:
> BC ok because of low impact

might still worth a minor right?

PR #1500, opened 2023-12-14
fix(recommend): update recommended-for-you model type

raed667:
When calling `getRecommendations` we need to make sure that the user doesn't have to pass `objectID` for the model `recommended-for-you`.

Otherwise this error will occur: 

```
Property 'objectID' is missing in type '{ indexName: string; model: "recommended-for-you"; fallbackParameters: RecommendSearchOptions | undefined; queryParameters: (RecommendSearchOptions & { ...; }) | undefined; maxRecommendations: number | undefined; }' but required in type 'RecommendationsQuery'.
```` 

The fix should be a patch since the model is not publicly available yet.

### DIFF
--- a/packages/recommend/src/types/RecommendationsQuery.ts
+++ b/packages/recommend/src/types/RecommendationsQuery.ts
@@ -1,4 +1,3 @@
-import { RecommendModel } from './RecommendModel';
 import { RecommendSearchOptions } from './RecommendSearchOptions';
 
 export type RecommendationsQuery = {
@@ -10,7 +9,7 @@ export type RecommendationsQuery = {
   /**
    * The name of the Recommendation model to use.
    */
-  readonly model: RecommendModel;
+  readonly model: 'related-products' | 'bought-together' | 'looking-similar';
 
   /**
    * The `objectID` of the item to get recommendations for.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by marialungu — APPROVED** [review]

File: packages/recommend/src/types/RecommendationsQuery.ts [hunk 1]
```diff
@@ -39,3 +39,18 @@ export type RecommendationsQuery = {
...
+ */
+export type RecommendationsQueryWithoutObjectID = Omit<
+  RecommendationsQuery,
+  'objectID' | 'model'
+> & {
+  /**
+   * The name of the Recommendation model to use.
+   */
+  readonly model: 'recommended-for-you';
+};
```
francoischalifour, line 56: We don't seem to need this `RecommendationsQueryWithoutObjectID` type. In `RecommendedForYouQuery` we can also omit `objectID`, and use `RecommendedForYouQuery` in `getRecommendations()`?

raed667: Good question, for `getRecommendedForYou` I also need to omit `model` from the parameters.

raed667: Addressed feedback in last commit

raed667:
Hey @francoischalifour , sorry i had some unpushed changes 😵‍💫 the PR should match what we talked about.

francoischalifour (approved):
Nice!

marialungu (approved):

PR #1504, opened 2024-01-22
chore: add-homepage-to-package.json

Adnan-Bacic:
in vscode for example, the homepage field will show when you hover over a library name in our own package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "version": "4.22.1",
   "private": true,
   "license": "MIT",
+  "homepage": "https://github.com/algolia/algoliasearch-client-javascript#readme",
   "workspaces": [
     "packages/*"
   ],


package.json
```diff
@@ -2,6 +2,7 @@
   "version": "4.24.0",
   "private": true,
   "license": "MIT",
+  "homepage": "https://github.com/algolia/algoliasearch-client-javascript#readme",
```
Haroenv, line 5: shouldn't this be in the package.json of `algoliasearch` instead of the root one? Also I wonder if it shouldn't be the documentation link instead, as the repo is already linked too

shortcuts: yes plus it's targeting v4, we don't release anything on this version except security/bug fixes

@Adnan-Bacic would you mind porting the PRs to https://github.com/algolia/api-clients-automation instead?

Adnan-Bacic: i dont quite understand why, but okay

millotp: This whole repository is generated, and the source is from https://github.com/algolia/api-clients-automation, once your other PR is merged it will be reflected here

Issue #1507, opened 2024-02-14
Recommend client is incorrectly sending Search user agents


sarahdayan:
## Actual

The Recommend client uses the same transporter utilities as the Search client, including the `createUserAgent` function. This causes it to set the "Algolia for JavaScript" user agent, which corresponds to Algolia Search, because `createUserAgent` [sets this value by default](https://github.com/algolia/algoliasearch-client-javascript/blob/252228ab7c5d5c94890885e8ce9bb9b072905756/packages/transporter/src/createUserAgent.ts#L5). That's likely because the transporter was initially designed solely for the Search API, before Recommend existed.

Additionally, the Recommend passes it its own version to `createUserAgent`, resulting in the Search client agent being sent with the Recommend client version.

## Expected

The Recommend client shouldn't send Algolia Search user agents at all.

## Suggested solution

The `createUserAgent` function shouldn't set "Algolia for JavaScript" by default. Instead, this agent should be added in the build of the Search client.

Haroenv:
Algolia for JavaScript is still relevant for recommend, no?

shortcuts:
> The `createUserAgent` function shouldn't set "Algolia for JavaScript" by default. Instead, this agent should be added in the build of the Search client.

IMHO this segment can stay because what differentiates the clients are those next segments https://github.com/algolia/algoliasearch-client-javascript/blob/252228ab7c5d5c94890885e8ce9bb9b072905756/packages/recommend/src/builds/node.ts#L39-L42 added in the init directly

Keeping the `Algolia for JavaScript` segment also allows us to track the global usage of the whole JavaScript API client

shortcuts:
Hey I think we can close this discussion now that things are cleared

PR #1508, opened 2024-02-20
fix(recommend): remove the search user-agent from requests [RECO-2154]

raed667:
## What

The createUserAgent function shouldn't set "Algolia for JavaScript" by default. Instead, this agent should be added in the build of the Search client.

## Why 

> The Recommend client uses the same transporter utilities as the Search client, including the createUserAgent function. This causes it to set the `Algolia for JavaScript` user agent, which corresponds to Algolia Search.

> The Recommend client shouldn't send Algolia Search user agents

## How 

Updated the function `createUserAgent` to accept an optional base UA as a second parameter. Recommend client can set that to an empty string, while the rest can still use the default value `Algolia for JavaScript`

### DIFF
--- a/packages/recommend/src/__tests__/recommend-client.test.ts
+++ b/packages/recommend/src/__tests__/recommend-client.test.ts
@@ -108,14 +108,12 @@ describe('recommend', () => {
     const client = recommend('appId', 'apiKey');
 
     if (testing.isBrowser()) {
-      expect(client.transporter.userAgent.value).toEqual(
-        `Algolia for JavaScript (${version}); Recommend (${version}); Browser`
-      );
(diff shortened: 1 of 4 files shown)


Haroenv:
If you look at #1507, I thought the conclusion was that "Algolia for JavaScript" isn't a search UA, but just a "JS client UA". Not sure this is needed

raed667:
Good catch @Haroenv i forgot about the Github issue 👀 i'll convert to draft for now and continue the discussion there.

shortcuts:
Same comments as in the issue for me too, the UA is the whole string being sent, `Algolia for JavaScript` is not scoped to the search client, the following segment defines the current client

PR #1522, opened 2024-04-16
feat(types): add `banners` in `renderingContent`

aymeric-giraudet:
These are types for the upcoming no-code Banner feature.

### DIFF
--- a/packages/client-search/src/types/Settings.ts
+++ b/packages/client-search/src/types/Settings.ts
@@ -358,6 +358,49 @@ export type Settings = {
         };
       };
     };
+    /**
+     * Defining UI widget configuration
+     */
+    readonly widgets?: {
+      /**
+       * Configuration for banners
+       */
+      readonly banners?: ReadonlyArray<{
+        /**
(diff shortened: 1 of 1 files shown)


taylorcjohnson (dismissed):
Approving since the failing tests don't appear to be related.

shortcuts (dismissed):
looks great, could you please reflect it in https://github.com/algolia/api-clients-automation/blob/b621988bce6f17f53bc12cc04d2f852d79b3cfee/specs/common/schemas/IndexSettings.yml#L983?

for the CI, it's super flaky so... you'll have to re-run it a bunch of times :(

aymeric-giraudet:
@shortcuts doesn't the Swagger spec file get generated itself too ?
If not we may do it manually but it should be done by those owning the backend code

shortcuts:
> @shortcuts doesn't the Swagger spec file get generated itself too ? If not we may do it manually but it should be done by those owning the backend code

no the spec is the entrypoint that generates everything

shortcuts (approved):

PR #1530, opened 2024-06-24
feat(types): add missing virtual attribute to Index type

kombucha:
This attribute is indeed present as part of the list indices payload, but is missing from the `Index` type.


Haroenv (approved):

shortcuts:
(i'm currently trying to fix the flaky ci, bear with me D:)

Issue #1540, opened 2024-08-21
Module types missing from manifest


Zach-Jaensch:
We have just upgraded from v4 to v5 (specifically v5.1.0) and ran into some issues when using the methods, primarily with the types. 
We noticed some issues like:
`Unsafe assignment of an error typed value.eslint[@typescript-eslint/no-unsafe-assignment](https://typescript-eslint.io/rules/no-unsafe-assignment)`
![image](https://github.com/user-attachments/assets/59660419-1dc1-4255-a7a2-cfbea275d4ba)

`Could not find a declaration file for module '@algolia/client-search'`
![image](https://github.com/user-attachments/assets/5d9476e3-edc3-4e2e-b6aa-26df0e41192c)

After some investigation, I when adding the types to the exports field in the manifest, the issues would be mostly solved
The client search package for example. 
```diff
  "exports": {
    ".": {
      "node": {
        "import": "./dist/client-search.esm.node.js",
        "module": "./dist/client-search.esm.node.js",
        "require": "./dist/client-search.cjs",
-       "default": "./dist/client-search.cjs"
+       "default": "./dist/client-search.cjs",
+       "types": "./dist/builds/node.d.ts"
      },
      "default": {
        "umd": "./dist/client-search.umd.js",
        "module": "./dist/client-search.esm.browser.js",
        "import": "./dist/client-search.esm.browser.js",
-       "default": "./dist/client-search.cjs"
+       "default": "./dist/client-search.cjs",
+       "types": "./dist/builds/browser.d.ts"
      }
    },
    "./src/*": "./src/*.ts",
    "./model": "./model/index.ts"
  },
```

It looks like all the packages need to be updated in a similar fashion which will allow the, otherwise when using the clients, we have to define the client list so (Note we patched the two packages to allow us to upgrade now) 
```typescript
import { algoliasearch } from "algoliasearch";

// These should not be needed 
import type { SearchClient } from "@algolia/client-search";
import type { RecommendClient } from "@algolia/recommend";

// this return type should be infered
export function getAlgoliaClient(): SearchClient {
  return algoliasearch(ALGOLIA_APP_ID, ALGOLIA_API_KEY);
}

// as above
export function getAlgoliaRecommendationClient(): RecommendClient {
  return algoliasearch(ALGOLIA_APP_ID, ALGOLIA_API_KEY).initRecommend();
}
```

PR #1541, opened 2024-08-21
fix: add types to manifest exports

Zach-Jaensch:
Apologies in advance if I have missed something.

fixes: #1540 

Type exports were missing from the manifest. I have verified locally that the types now work, however have noticed that the inferred return type from algoliasearch is still failing, possibly due to the use of `*.ts` files from the other packages. This may require changes to the build config. 

I can create a minimum repo to help with testing in the morning if need be.

### DIFF
--- a/packages/algoliasearch/package.json
+++ b/packages/algoliasearch/package.json
@@ -16,13 +16,15 @@
         "import": "./dist/algoliasearch.esm.node.js",
         "module": "./dist/algoliasearch.esm.node.js",
         "require": "./dist/algoliasearch.cjs",
-        "default": "./dist/algoliasearch.cjs"
+        "default": "./dist/algoliasearch.cjs",
+        "types": "./dist/algoliasearch/builds/node.d.ts"
       },
       "default": {
         "umd": "./dist/algoliasearch.umd.js",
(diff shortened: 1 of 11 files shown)


shortcuts:
Hey, thanks for opening the PR, I was actually answering your issue!

The changes look good but needs to be ported to https://github.com/algolia/api-clients-automation, more specifically in https://github.com/algolia/api-clients-automation/blob/main/templates/javascript/clients/package.mustache, as those API clients are generated.

Would you like me to move your PR there, or do you have time to do it? thanks!

Zach-Jaensch:
Hey, ah I figured I would have missed something. I'm Australian based so about to call it for the night. Happy to re-do it in the morning, also just as happy for you to close this and port it over. I'll take a look when I start up in the morning. Thanks!

shortcuts:
> Hey, ah I figured I would have missed something. I'm Australian based so about to call it for the night. Happy to re-do it in the morning, also just as happy for you to close this and port it over. I'll take a look when I start up in the morning. Thanks!

Don't worry, enjoy a relaxing night I'll port the changes :D Thanks again for contributing

Issue #1537, opened 2024-08-16
No clear migration guide for v5 and no explicit typing


WavyWalk:
We want to migrate to v4 from v5, but migrating even this snippet is problematic:
```typescript
const object = await algoliasearch(algoliaApplicationId, algoliaApiKey)
      .initIndex(productsAlgoliaIndex)
      .getObjects<AlgoliaRecord>(productIds)
```

I see in source code that getObjects exist in v5 but one can't pass index and it's not in docs, and there's no index in types.

Also types are not straightforward (I get that it's generated but still), so rules like @typescript-eslint/no-unsafe-assignment fail because of any's in unions.

As well not sure how to proceed with now failing type checks for react instantsearch because things like:
hitsPerPage, analyticsTags on `<Configure /> `reference the types of v5 which do not have it.

Do you recommend to migrate now or wait till there will be a proper migration guide and types will be improved?

OutdatedGuy:
You need to update your code using [this documentation](https://www.algolia.com/doc/libraries/javascript/v5/upgrade/).

Before:
```ts
import algoliasearch from "algoliasearch";

const object = await algoliasearch(algoliaApplicationId, algoliaApiKey)
  .initIndex(productsAlgoliaIndex)
  .getObjects<AlgoliaRecord>(productIds);
```

After:
```ts
import { algoliasearch } from "algoliasearch";

const object = await algoliasearch(algoliaApplicationId, algoliaApiKey)
  .getObjects<AlgoliaRecord>({
    requests: [
      {
        indexName: productsAlgoliaIndex,
        objectID: productId1,
      },
      {
        indexName: productsAlgoliaIndex,
        objectID: productId2,
      },
      // ...
    ],
  });
```

shortcuts:
Hey @WavyWalk thanks for using the v5 client already, and sorry for the lack of documentation, we are currently working on providing snippets and guides! (and thanks @OutdatedGuy for the help :D)

> I see in source code that getObjects exist in v5 but one can't pass index and it's not in docs, and there's no index in types.

The `initIndex` of the previous version was kind of magical syntactic sugar (and not suited for generated clients too), as [mentioned here](https://www.algolia.com/doc/libraries/javascript/v5/upgrade/#removal-of-initindex), every methods are now available at the root of the client, if you were using one via `initIndex` before, it will now expect an `indexName` parameter if it's a single-index method, or an `indexName` in the `requests` if it's a multi-index method

[...]

PR #1546, opened 2024-08-31
Added generic type to getObject method

hirotomoyamada:
Related #1537 

Hi, thank you for developing a very nice library.

As mentioned in this issue, the type of the `getObject` function is incorrect.


hirotomoyamada:
Fixed #3656

PR #1573, opened 2024-12-14
feat(client-search): removed node apis from worker build

### LINKED ISSUE #1567: [bug]: Unable to use search client in cloudflare workers: Cannot bundle Node.js built-in "crypto"
### Description

Using version `5.10.2` we cannot bundle algolia search client for cloudflare workers since it uses Node built ins.

We tried updating from `5.2.1`, which still works fine.

### Client

All

### Version

5.10.2

### Relevant log output

_No response_

TorbjornHoltmon:
* Removed node specific imports from the worker build of client search, replaced them with the web crypto API
* Added test to verify that the result is the same between node crypto and web crypto

This pull request makes `generateSecuredApiKey` async, this can cause breaking behavior.
Closes #1567


shortcuts:
Hey, thanks a lot for your PR! The changes looks really nice, this needs to however be ported to https://github.com/algolia/api-clients-automation as this repository content is fully generated.

I'll move some of your commits there and let you know once it's good :)

shortcuts:
Hey again, i've ported changes to https://github.com/algolia/api-clients-automation/pull/4249, also made it to kind of avoid breaking changes so it's a worker specific fix instead

PR #1586, opened 2025-05-20
feat: add transformation bridge methods

shortcuts:
https://algolia.atlassian.net/browse/DI-3785

this implements the bridge methods in algoliasearch to the ingestion api, in order to reduce friction with the transformation pipeline indexing

related: https://github.com/algolia/api-clients-automation/pull/4852


packages/algoliasearch/src/types/Ingestion.ts
```diff
@@ -0,0 +1,252 @@
...
+  };
+  /**
+   * When provided, the push operation will be synchronous and the API will wait for the ingestion to be finished before responding.
+   */
+  readonly watch?: boolean;
+};
+
+export type PushOptions = Pick<PushProps, 'watch'>;
+
+export const createIngestionClient = (
```
Haroenv, line 92: this is in a types folder but contains real code, is that intentional?

packages/algoliasearch/src/types/Ingestion.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    const action = autoGenerateObjectIDIfNotExist
+      ? BatchActionEnum.AddObject
+      : BatchActionEnum.UpdateObject;
+
+    /* eslint functional/immutable-data: "off" */
+    return await client.push(
+      {
+        indexName,
+        pushTaskPayload: { action, records: objects },
```
Haroenv, line 212: if I understand it right, there's an endpoint now that is generically push without needing a specific id?

Haroenv (commented):
looks like a good idea. Maybe unfortunate that this would be a lot of changes to have as a simple option to saveObject etc. as that would be simpler for the user.

shortcuts:
> looks like a good idea. Maybe unfortunate that this would be a lot of changes to have as a simple option to saveObject etc. as that would be simpler for the user.

The response type differs so we can't really do that without impacting everyone, and I think the discoverability is better with a standalone method rather than an option, wdyt?

packages/algoliasearch/src/types/Ingestion.ts
shortcuts: Exactly! We just added a new endpoint that does the lookup for the customer based on the indexName

packages/algoliasearch/src/types/Ingestion.ts
shortcuts: Indeed I forgot to move it to its own file

packages/algoliasearch/src/ingestion.ts
```diff
@@ -0,0 +1,167 @@
...
+  PushOptions,
+  PushProps,
+  TransformationOptions,
+  WatchResponse,
+} from './types';
+
+export function createIngestionClient(
+  options: SearchClientOptions & ClientTransporterOptions & TransformationOptions
+): IngestionClient {
+  if (!options || !options.transformation || !options.transformation.region) {
```
millotp, line 23: maybe check that the value is either `us` or `eu`

packages/algoliasearch/src/ingestion.ts
```diff
@@ -0,0 +1,167 @@
...
+    throw new Error('`region` must be provided when leveraging the transformation pipeline');
+  }
+
+  const appId = options.appId;
+
+  const auth = createAuth(
+    options.authMode !== undefined ? options.authMode : AuthMode.WithinHeaders,
```
millotp, line 30: ingestion only supports header auth mode

packages/algoliasearch/src/ingestion.ts
```diff
@@ -0,0 +1,167 @@
...
+      {
+        url: `data.${options.transformation.region}.algolia.com`,
+        accept: CallEnum.ReadWrite,
+        protocol: 'https',
+      },
+    ],
+    ...options,
+    headers: {
+      ...auth.headers(),
+      ...{ 'content-type': 'application/x-www-form-urlencoded' },
```
millotp, line 46: is this volontary ? we only support application/json

packages/algoliasearch/src/ingestion.ts
```diff
@@ -0,0 +1,167 @@
...
+      ...auth.queryParameters(),
+      ...options.queryParameters,
+    },
+  });
+
+  return {
+    transporter,
+    appId,
+    addAlgoliaAgent(segment: string, version?: string): void {
+      transporter.userAgent.add({ segment, version });
```
millotp, line 59: can you also add `Ingestion;` so that it also count in ingestion metrics ?

Haroenv (commented):
idea works for me, details should be approved by ingestion or data team :)

packages/algoliasearch/src/ingestion.ts
shortcuts: ah ty I forgot to match what we had in the generated clients here

packages/algoliasearch/src/ingestion.ts
shortcuts: good catch

PR #1598, opened 2025-11-06
fix: always delete temporary index in replaceAllObjects

millotp:
If there is an error during the `saveObjects` or `move` in `replaceAllObjects`, it will leave behind the temporary index, and count in the index limit of the app.

Add a clean up step that will be delete the index in case of failure.


shortcuts (dismissed):
gg!

PR #1603, opened 2026-02-06
feat(requester-node-http): add gzip compression for requests and response decompression

Fluf22:
When the accept-encoding header includes gzip, request bodies are compressed with zlib.gzipSync() and a content-encoding: gzip header is added. Responses with content-encoding: gzip are decompressed via zlib.createGunzip(). Decompression errors resolve gracefully following the existing error handling pattern (never reject).

Includes 8 comprehensive tests covering compression, decompression, error handling, multi-value headers, and full round-trip scenarios.

### DIFF
--- a/packages/requester-node-http/src/createNodeHttpRequester.ts
+++ b/packages/requester-node-http/src/createNodeHttpRequester.ts
@@ -140,14 +140,26 @@ export function createNodeHttpRequester({
           responseTimeout = createTimeout(request.responseTimeout, 'Socket timeout');
         });
 
-        if (request.data !== undefined) {
-          const body = shouldCompress ? zlib.gzipSync(request.data) : request.data;
+        if (request.data !== undefined && shouldCompress) {
+          zlib.gzip(request.data, (error, compressedBody) => {
+            if (error) {
+              onError(error);
 
-          req.setHeader('content-length', Buffer.byteLength(body));
-          req.write(body);
+              return;
+            }
+
+            req.setHeader('content-length', compressedBody.byteLength);
+            req.write(compressedBody);
+            req.end();
+          });
+        } else {
+          if (request.data !== undefined) {
+            req.setHeader('content-length', Buffer.byteLength(request.data));
+            req.write(request.data);
+          }
+
+          req.end();
         }
-
-        req.end();
       });
     },
 


packages/requester-node-http/src/createNodeHttpRequester.ts
```diff
@@ -80,24 +117,21 @@ export function createNodeHttpRequester({
...
 
         req.once('response', () => {
           clearTimeout(connectTimeout);
           responseTimeout = createTimeout(request.responseTimeout, 'Socket timeout');
         });
 
         if (request.data !== undefined) {
-          req.write(request.data);
+          if (shouldCompress) {
+            req.write(zlib.gzipSync(request.data));
```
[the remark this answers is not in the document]
Fluf22: I believe the default is 3

https://nodejs.org/api/zlib.html

packages/requester-node-http/src/createNodeHttpRequester.ts
```diff
@@ -42,30 +49,60 @@ export function createNodeHttpRequester({
...
+            const gunzip = zlib.createGunzip();
+
+            response.pipe(gunzip);
+
+            gunzip.on('data', onData);
+            gunzip.on('end', onEnd);
+            gunzip.on('error', onError);
+          } else {
+            response.on('data', onData);
+            response.on('end', onEnd);
```
Ant-hem: Shouldn't we have `response.on('error', onError);` ? Same I am not fluent with nodejs so please dismiss if not on point.

packages/requester-node-http/src/createNodeHttpRequester.ts
```diff
@@ -42,30 +49,60 @@ export function createNodeHttpRequester({
...
+          clearTimeout(responseTimeout as NodeJS.Timeout);
+        };
+
+        const onError = (error: Error): void => {
+          clearTimeouts();
+          resolve({ status: 0, content: error.message, isTimedOut: false });
+        };
+
         const req = (url.protocol === 'https:' ? https : http).request(options, response => {
+          const contentEncoding = response.headers['content-encoding'];
```
Ant-hem: Maybe add a `tolower` ?

packages/requester-node-http/src/createNodeHttpRequester.ts
Fluf22: 🙊 That's a miss indeed, adding it

packages/requester-node-http/src/createNodeHttpRequester.ts
```diff
@@ -42,36 +51,78 @@ export function createNodeHttpRequester({
...
+          response.on('error', onError);
+
+          if (isGzipResponse) {
+            gunzip = zlib.createGunzip();
+
+            response.pipe(gunzip);
+
+            gunzip.on('data', onData);
+            gunzip.on('end', onEnd);
+            gunzip.on('error', onError);
```
samouss: I'm a bit rusty in JavaScript so please disregard if not relevant: Why do we have to listen on the gzip object AND the response? We pipe the gzip object on the response so why don't we listen once on the response instead of branching?

Ant-hem (approved):
> @Ant-hem @samouss any last comments?

All good on my side! Would be perfect to have a review/approval from somebody fluent in JS tho :)

packages/requester-node-http/src/createNodeHttpRequester.ts
samouss: Ok. I thought the `pipe()` was returning a new stream which we could listen once.

```js
if (isGzipResponse) {
  gunzip = zlib.createGunzip();
  response = response.pipe(gunzip);
}

response.on(...)
```

packages/requester-node-http/src/createNodeHttpRequester.ts
```diff
@@ -33,6 +34,14 @@ export function createNodeHttpRequester({
 
         const path = url.query === null ? url.pathname : `${url.pathname}?${url.query}`;
 
+        const COMPRESSION_THRESHOLD = 750;
```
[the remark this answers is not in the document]
Fluf22: I'll add one in the doc I'm making for compression implem in the new clients

packages/requester-node-http/src/createNodeHttpRequester.ts
Fluf22: Doing that, we would really be listening on the destination stream (here, `gunzip`), which is returned when you do `.pipe(destination)`, so any error happening in the `response` would be shadowed

packages/requester-node-http/src/createNodeHttpRequester.ts
Fluf22: Ah shit, I just noticed you were talking about the header key, not value

PR #1612, opened 2026-09-07
fix(requester-node-http): remove listeners after request settles

### LINKED ISSUE #1570: [bug]: Memory leak in Node requester
### Description

We got reports of memory leaks from users of React InstantSearch with Next.js.

It's easy to reproduce, with an `index.mjs` file : 
```js
import process from 'process';

import { algoliasearch } from 'algoliasearch';

process.stdin.resume();

const client = algoliasearch('latency', '6be0576ff61c053d5f9a3225e2a90f76');

for (let i = 0; i < 1000; i++) {
  client.search({ requests: [{ indexName: 'instant_search' }] });
}
```

Then run `node --inspect index.mjs`, in Chrome open `chrome://inspect` and select `index.mjs`.

In the memory tab, the VM instance heap size hovers around 90MB without ever going down.

image

All responses are retained by the error handler.

I tried patching it by adding `req.removeAllListeners()` as well as `response.removeAllListeners()` in this `end` handler : https://github.com/algolia/algoliasearch-client-javascript/blob/36ccbc9799823ccb4868460292f921d1929941e6/packages/requester-node-http/src/createHttpRequester.ts#L60-L69

It's not perfect but it now settles around 20MB. Some unnecessary values are still there but maybe this will properly get GC'd if needed.
image

I told them to use `requester-fetch` instead as it's now supported by recent versions of Node.js. Maybe we could make that the default ? From my own benchmarks it seems better than the Node requester currently but with my proposed fix it would make the Node requester more performant than the fetch one 😅 (with undici)

### Client

All

### Version

5.15.0

### Relevant log output

_No response_

Steve0x2a:
`createHttpRequester` attached `data` / `end` / `error` listeners on each Node `http` response (and `error` on the request) and never tore them down. With the keep-alive agent those emitters stay reachable, so the handlers keep response buffers alive across repeated `client.search()` calls.

On success, error, and timeout, this now calls `removeAllListeners()` on the request and response so those handlers cannot retain the settled exchange.

Thanks @aymeric-giraudet for the report and the `removeAllListeners` workaround.

Fixes #1570


Haroenv:
This PR looks good, you can recreate it on https://github.com/algolia/api-clients-automation :) in the relevant template file

